PR #279, opened 2022-05-27
Add wasm output packaging as shutter-crypto NPM package

ulope:
This uses webpack to generate a NPM package `shutter-crypto` from the wasm target.

Also adds size optimization of the WASM output by using wasm-opt.


schmir-at-bb (approved):
LGTM

Issue #269, opened 2022-05-06
Reevaluate EpochID encoding


ezdac:
Following up on https://github.com/shutter-network/rolling-shutter/pull/239 @jannikluhn and I had the discussion whether 
the `<uint32 block_number><uint32 batch index>` coding of the EpochID is necessary.
We think that the EpochID can be any arbitrary number decided upon by the collator, as long as they stay unique (no reuse).
We did not see disadvantages to simply set the `EpochID` as the L2 `BatchIndex`, but thought this makes things easier.

Any objections? Why was the block number encoded in the epoch-id?

jannikluhn:
Done in 291.

PR #291, opened 2022-06-03
Pure Epoch IDs

jannikluhn:
The goal of this PR is to make epoch ids arbitrary 32 byte strings, instead of encoding additional information (like a block number). This makes the protocol more flexible, in particular it supports proposal ids for snapshot shutter.

The implicitly encoded block number was used in a couple of ways:

- To identify responsible keyper set and eon in decryption trigger messages. This has been replaced by an explicit block number
- To identify responsible keyper set and eon in key share and decryption key messages. This has been replaced by an explicit eon number (note that we assume it's unique).
- To identify l1 block number to be used in the next batch in the batch production cycle. This has been replaced by an explicit block number stored along the epoch id.


jannikluhn:
closes #269

schmir-at-bb (approved):
LGTM

PR #292, opened 2022-06-03
Limit total batch data size

### LINKED ISSUE #281: Collator-BatchHandler: calldata limit checks
While we check for the gas-limit, we do not yet have a calldata restriction in the validation method for incoming transactions.
Once Sequencer-Optimism is implemented, replicate the calldata tracking / validation from the optimism state transition function and use it in the `Batch` struct, similar to the gas-pool tracking.

jannikluhn:
Via a simple hardcoded limit.

Closes #281

### DIFF
--- a/rolling-shutter/collator/batch/batch.go
+++ b/rolling-shutter/collator/batch/batch.go
@@ -16,7 +16,10 @@ import (
 	"github.com/shutter-network/rolling-shutter/rolling-shutter/medley/epochid"
 )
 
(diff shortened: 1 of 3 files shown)


ezdac (approved):
Looks good!

But is there a spec somewhere or is this just a placeholder value until the implementation is known?
I couldn't find this check in current cannon or optimism state-transition so I can't comment on the `8*1024` bytes and whether the whole byte-coded transaction should be counted.

PR #304, opened 2022-06-30
Fix L1/L2 block numbers

### LINKED ISSUE #301: Make activation block numbers refer to mainchain
The activation block number is a main chain block number that defines when a configuration should come into effect. However, in the contracts we often compare it to `block.number` which refers to the L2 block number. We should get rid of these checks.

jannikluhn:
Closes #301 

The contracts are deployed on L2, but we use the L1 block number as a clock. Two changes:

1) Collator: Expect the contracts to be deployed on the rollup (to which we're connected via `SequencerURL`). Get block numbers for batches from `EthereumURL`.
2) Keyper: Add an L2 node URL as additional config option. Get contracts from there and block numbers from `EthereumURL`. For `BlockSeen` messages, fetch block numbers from `EthereumURL` (instead of using `NextBlock` from the chain observer, which observes the L2 chain).

For simplicity of the testing infrastructure, we use the same node as L1 and L2.

### DIFF
--- a/rolling-shutter/keyper/keyper.go
+++ b/rolling-shutter/keyper/keyper.go
@@ -28,6 +28,7 @@ import (
 	"github.com/shutter-network/rolling-shutter/rolling-shutter/keyper/fx"
 	"github.com/shutter-network/rolling-shutter/rolling-shutter/keyper/kprdb"
 	"github.com/shutter-network/rolling-shutter/rolling-shutter/keyper/kproapi"
+	"github.com/shutter-network/rolling-shutter/rolling-shutter/medley"
 	"github.com/shutter-network/rolling-shutter/rolling-shutter/medley/eventsyncer"
 	"github.com/shutter-network/rolling-shutter/rolling-shutter/p2p"
 	"github.com/shutter-network/rolling-shutter/rolling-shutter/shdb"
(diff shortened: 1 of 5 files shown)


rolling-shutter/keyper/keyper.go
```diff
@@ -246,21 +253,22 @@ func (kpr *keyper) handleOnChainChanges(ctx context.Context, tx pgx.Tx) error {
...
 func (kpr *keyper) sendNewBlockSeen(ctx context.Context, tx pgx.Tx) error {
-	qc := commondb.New(tx)
 	q := kprdb.New(tx)
 	lastBlock, err := q.GetLastBlockSeen(ctx)
 	if err != nil {
 		return err
 	}
-	nextBlock, err := qc.GetNextBlockNumber(ctx)
+	currentBlock, err := medley.Retry(ctx, func() (uint64, error) {
+		return kpr.l1Client.BlockNumber(ctx)
```
schmir-at-bb, line 262: This is running inside an active database transaction. We should not do any HTTP requests during that.

schmir-at-bb:
I do wonder why we use L1 block number as a clock if all of the contracts are still deployed on L2. Can you explain the reasoning behind that?

rolling-shutter/keyper/keyper.go
jannikluhn: fixed

jannikluhn:
The L1 block number is safer to use because it's not under the control of the sequencer. E.g., the sequencer could try to stall collator updates by producing blocks at a smaller rate.

For the contracts, it doesn't really matter if the activation blocks are specified as L1 or L2 blocks, for them it can be just an abstract measure of time.

schmir-at-bb:
ok. thanks for the explanation.

PR #309, opened 2022-07-14
Some shcrypto js changes

jannikluhn:
- Users have to specify URL/path to the WASM file. This seems to be the most flexible, but requires some manual work on the users side. On node could look something like this: `await shutterCrypto.init("./node_modules/@shutter-network/shutter-crypto")`. For frontends, users can first `ln -s node_modules/@shutter-network/shutter-crypto/dist/shutter-crypto.wasm public` and then initialize with `await shutterCrypto.init("shutter-crypto.wasm")`. This seems acceptable to me. I failed to find a more automatic way to do it.
- functions are explicitly async
- errors returned from WASM are thrown as exceptions
- return values are converted from hex to Uint8Array (same format as inputs are expected)

Builds on top of #308

### DIFF
--- a/js/shutter-crypto/Makefile
+++ b/js/shutter-crypto/Makefile
@@ -6,9 +6,13 @@ TINYGO    ?= tinygo
 
 build: derive
 	${NPX} webpack build
+	cp -L derived/shutter-crypto.wasm dist/
 
 publish: build
 	${NPM} publish --access=public --tag=beta
+
+pack: build
+	${NPM} pack
 
 derive: derived/shutter-crypto.wasm derived/wasm_exec.js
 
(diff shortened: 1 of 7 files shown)


ulope:
Yeah that's probably the best solution. 

IMO for node we could still automatically detect the path. That would improve usability in that case (also I'm not sure that assuming the project root is the current directory is going to be valid in all cases).

ulope:
I re-implemented the automatic detection of the wasm path for node.

Also this fixes a race when initializing the WASM/GO runtime which could lead to erroneous `You need to consume the 'shutterCrypto.init()' promise before using the module functions.`.

PR #312, opened 2022-07-29
`batch` package split and threadsafety improvements

### LINKED ISSUE #286: Collator batch handler: synchronize epoch start with latest batch inclusion
Currently a thread in the collator will continuously trigger the start of new epochs with a time ticker every X seconds.
That interval should only be the minimum time between epochs though:

We need to wait on the successful inclusion of the previous epoch in the l2 chain state before we can validate and include transactions for the next batch. This is because the validation of pending transactions depends on the state-changes caused by the previous batches' decrypted transactions.

Starting a new epoch is dependent on the validated pending batches' transactions (for the Decryption trigger), so we should initiate sending the  trigger only when:
1) the ticker time interval since the last epoch start was reached
2) **AND** the previous batch-transaction has been successfully sent, confirmed and is included in the latest state of the l2 chain

ezdac:
*This PR is part of a bigger feature implementation ( #286 )
Please go through through the review commit by commit and keep in mind that a series of PRs
will follow so that ultimately the `feature/collator-batch-handler` branch can be merged to `main` to fix #286.*

This PR includes:
- refactoring of the `batch` into sub-packages
- threadsafety fixes for several classes
- add an JSON-RPC endpoint to the mock eth-server implementation

### DIFF
--- a/rolling-shutter/collator/batchhandler/batch/batch.go
+++ b/rolling-shutter/collator/batchhandler/batch/batch.go
@@ -10,9 +10,12 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/pkg/errors"
 	txtypes "github.com/shutter-network/txtypes/types"
 
+	"github.com/shutter-network/rolling-shutter/rolling-shutter/collator/batchhandler/sequencer"
+	"github.com/shutter-network/rolling-shutter/rolling-shutter/collator/batchhandler/transaction"
 	"github.com/shutter-network/rolling-shutter/rolling-shutter/medley/epochid"
 )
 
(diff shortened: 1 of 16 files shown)


rolling-shutter/collator/batchhandler/sequencer/statecache.go
```diff
@@ -27,34 +30,104 @@ type EthClient interface {
 	NonceAt(context.Context, common.Address, *big.Int) (uint64, error)
 }
 
-func NewChainBatchCache(client EthClient, atBlockNumber *big.Int) *ChainBatchCache {
-	return &ChainBatchCache{
+type countedMux struct {
+	mux *sync.RWMutex
+	cnt uint32
```
jannikluhn, line 35: Maybe clarify that we use the same counter for `Lock/Unlock` and `RLock/RUnlock`, at least to me this was confusing (and not self-evident that it's ok to do so).

rolling-shutter/collator/batchhandler/sequencer/statecache.go
```diff
@@ -27,34 +30,104 @@ type EthClient interface {
...
 
 	Client        EthClient
 	AtBlockNumber *big.Int
 }
 
 // GetBalance polls and caches the state of account `a` balance at the
 // block number ChainBatchCache.AtBlockNumber.
-func (c *ChainBatchCache) GetBalance(ctx context.Context, a common.Address) (*big.Int, error) {
+func (c *Cached) GetBalance(ctx context.Context, a common.Address) (*big.Int, error) {
+	// write lock because we eventually write to the dict
```
jannikluhn, line 124: why not make `getBalance` do this? Same for `subBalance` etc.

jannikluhn (approved):

rolling-shutter/collator/batchhandler/sequencer/statecache.go
```diff
@@ -0,0 +1,224 @@
...
+type countedMux struct {
+	mux *sync.RWMutex
+	cnt uint32
+}
+
+func NewKeyedMutex[T comparable]() *KeyedMutex[T] {
+	return &KeyedMutex[T]{mu: sync.RWMutex{}, muxs: make(map[T]*countedMux)}
+}
+
+type KeyedMutex[T comparable] struct {
```
jannikluhn, line 42: nice utility, and yay generics!

rolling-shutter/collator/batchhandler/sequencer/statecache.go
```diff
@@ -27,34 +30,104 @@ type EthClient interface {
...
+	mu   sync.RWMutex
+	muxs map[T]*countedMux
+}
+
+func (ml *KeyedMutex[T]) call(method string, v T, delta int) {
```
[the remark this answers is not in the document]
ezdac: Ah good idea, this will make the code more readable and less magic

rolling-shutter/collator/batchhandler/sequencer/statecache.go
ezdac: Will document it.
I also was not sure initially, but this counter functions as a reference counter to the mutex object.
It tracks general usage of the mutex in order to not keep dangling objects hanging in the map
indefinitely, and to instantiate a new one when the counter went to 0 before.
The object itself decides about read/write access separation.

rolling-shutter/collator/batchhandler/sequencer/statecache.go
ezdac: This doesn't work, because subBalance calls `getBalance`.
If the method in the lowest layer acquires the mutex you would get nested write-mutex calls, which will deadlock

rolling-shutter/collator/batchhandler/batch/batch.go
```diff
@@ -61,38 +64,36 @@ func CalculatePriorityFee(tx *txtypes.Transaction, baseFee *big.Int) *big.Int {
 	return new(big.Int).Mul(priorityFeeGasPrice, new(big.Int).SetUint64(tx.Gas()))
 }
 
-func NewPendingTransaction(signer txtypes.Signer, txBytes []byte, receiveTime time.Time) (*PendingTransaction, error) {
-	var tx txtypes.Transaction
-	err := tx.UnmarshalBinary(txBytes)
+func NewCachedPendingBatch(
```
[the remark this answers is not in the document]
ezdac: I haven't found any convention about this. But I agree with you and simply haven't put any thought into this.

rolling-shutter/collator/batchhandler/sequencer/statecache.go
```diff
@@ -0,0 +1,224 @@
...
+	var err error
+
+	bal, exists := c.balances[a]
+	if !exists {
+		bal, err = c.Client.BalanceAt(ctx, a, c.AtBlockNumber)
+		if err != nil {
+			return nil, err
+		}
+		c.balances[a] = bal
+	}
```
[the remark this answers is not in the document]
ezdac: The private method itself is not threadsafe, but the public interface to it guards the whole function call with a write lock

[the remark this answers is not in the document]
ezdac: Removed the threadsafety features of that class alltogether

PR #313, opened 2022-07-29
Add event observer classes

### LINKED ISSUE #286: Collator batch handler: synchronize epoch start with latest batch inclusion

ezdac:
*This PR is part of a bigger feature implementation ( #286 )
Please go through through the review commit by commit and keep in mind that a series of PRs
will follow so that ultimately the `feature/collator-batch-handler` branch can be merged to `main` to fix #286.*

This PR includes:
- addition of a generic observer-pattern implementation (`Broker`)
- addition of a specific broker for epoch-ticks - note this class still has a bug

### DIFF
--- a/rolling-shutter/medley/broker.go
+++ b/rolling-shutter/medley/broker.go
@@ -6,7 +6,7 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-// `Broker` allows to distribute a value to multiple
+// Broker allows to distribute a value to multiple
 // receiving channels, so that all channels receive the same value.
 // This is different to the Fan-Out pattern where multiple channels
 // are listening for a send on one channel, but only the first channel
@@ -32,24 +32,23 @@ type Broker[T any] struct {
 	unsubscribe          chan chan T
 	running              bool
 	bufferSizeSubscriber int
-	blockingSend         bool
+	nonBlockingSend      bool
 }
 
-func NewBroker[T any](buffer int, blocking bool) *Broker[T] {
+func NewBroker[T any](buffer int, nonBlocking bool) *Broker[T] {
 	return &Broker[T]{
-		stop:                 make(chan struct{}),
-		publish:              make(chan T, 1),
-		subscribe:            make(chan chan T),
-		unsubscribe:          make(chan chan T),
-		bufferSizeSubscriber: buffer,
-		blockingSend:         blocking,
-		running:              false,
+		stop:            make(chan struct{}),
+		publish:         make(chan T, 1),
+		subscribe:       make(chan chan T),
+		unsubscribe:     make(chan chan T),
+		nonBlockingSend: nonBlocking,
+		running:         false,
 	}
 }
 
-// `Start` starts the Broker's internal loop that processes
+// Start starts the Broker's internal loop that processes
 // the subscription and publish operations.
-// `initialSubscriptions` will be registered
+// initialSubscriptions will be registered
 // before the loop is started.
 // This is useful when the channels must not miss a published value
 // after the start or when the subscribers buffer size
@@ -85,7 +84,7 @@ func (b *Broker[T]) Start(initialSubscriptions ...chan T) {
 				close(msgCh)
 				log.Debug().Int("num-subscribers", len(subscriptions)).Msg("Unsubscribed")
 			case msg := <-b.publish:
-				if b.blockingSend {
+				if b.nonBlockingSend {
 					// use blocking send to make sure all subscribers
 					// got the value,
 					// could block the whole broker (+ successive Publish() calls) if a receiver
@@ -111,7 +110,7 @@ func (b *Broker[T]) Start(initialSubscriptions ...chan T) {
 	<-started
 }
 
-// `Stop` will close the internal stop channel
+// Stop will close the internal stop channel
 // and will stop the Brokers internal loop
 // as soon as the closed stop channel is selected.
 // A call to Stop() will also cause the subscribed
@@ -120,7 +119,7 @@ func (b *Broker[_]) Stop() {
 	close(b.stop)
 }
 
-// `Subscribe` registers the channel to the internal
+// Subscribe registers the channel to the internal
 // send operation. It will receive a published value
 // as soon as the other registered subscribers in line
 // received the value (in the blockingSend case),
@@ -134,7 +133,7 @@ func (b *Broker[T]) Subscribe() chan T {
 	return channel
 }
 
-// `Unsubscribe` deregisters the subscribed channel
+// Unsubscribe deregisters the subscribed channel
 // from the internal send operation and close the
 // subscribed channel.
 // This is a noop when the loop is not running.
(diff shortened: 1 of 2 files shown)


rolling-shutter/medley/broker.go
```diff
@@ -0,0 +1,184 @@
...
+// the subscribers buffer size - but if a subscriber is blocked forever,
+// this will only work until this subscribers buffer is full.
+type Broker[T any] struct {
+	mux                  sync.RWMutex
+	stop                 chan struct{}
+	publish              chan T
+	subscribe            chan chan T
+	unsubscribe          chan chan T
+	running              bool
+	bufferSizeSubscriber int
```
jannikluhn, line 34: `subscriberBufferSize` seems like a better name, the current one confused me until I looked at the implementation

rolling-shutter/medley/broker.go
```diff
@@ -0,0 +1,184 @@
+package medley
+
+import (
+	"sync"
+
+	"github.com/rs/zerolog/log"
+)
+
+// `Broker` allows to distribute a value to multiple
```
jannikluhn, line 9: Why the backticks? I think so far I've only seen unquoted names in godoc comments.

rolling-shutter/medley/broker.go
```diff
@@ -0,0 +1,184 @@
+package medley
+
+import (
+	"sync"
+
+	"github.com/rs/zerolog/log"
```
jannikluhn, line 6: Don't really have an opinion on which logging package to use, but I think Schmir has looked into it. Any comments, @schmir-at-bb ?

rolling-shutter/medley/broker.go
```diff
@@ -0,0 +1,184 @@
...
+
+// `Start` starts the Broker's internal loop that processes
+// the subscription and publish operations.
+// `initialSubscriptions` will be registered
+// before the loop is started.
+// This is useful when the channels must not miss a published value
+// after the start or when the subscribers buffer size
+// should diverge from the default value `bufferSizeSubscriber`
+// set in the constructor.
+func (b *Broker[T]) Start(initialSubscriptions ...chan T) {
```
jannikluhn, line 58: `initialSubscriptions` might have a different capacity than channels created with `Subscribe` which could lead to unexpected behavior. How about

- changing `Subscribe` to take a channel as a parameter (and thus get rid of `bufferSizeSubscriber`) or
- in `Start` take a parameter `numInitialSubscriptions` and creating and returning the channels with the expected size

?

rolling-shutter/medley/broker.go
```diff
@@ -0,0 +1,184 @@
...
+// this will only work until this subscribers buffer is full.
+type Broker[T any] struct {
+	mux                  sync.RWMutex
+	stop                 chan struct{}
+	publish              chan T
+	subscribe            chan chan T
+	unsubscribe          chan chan T
+	running              bool
+	bufferSizeSubscriber int
+	blockingSend         bool
```
jannikluhn, line 35: Do we expect to need non-blocking send? It seems dangerous because not only could some receivers not receive all values, but some values might be lost completely. And initializing `NewBroker` with `false` instead of `true` seems like a very easy to miss and hard to find bug. So if we don't have a clear use case already, I'd suggest to simply remove it.

rolling-shutter/medley/broker.go
```diff
@@ -0,0 +1,184 @@
...
+		return
+	}
+	b.mux.RUnlock()
+	for {
+		select {
+		case <-b.stop:
+			return
+		case b.unsubscribe <- channel:
+			return
+		case <-channel:
```
jannikluhn, line 156: I don't understand this, what does this do?

jannikluhn (approved):

rolling-shutter/medley/broker.go
ezdac: He introduced it somewhere else as well. I quite like this one

rolling-shutter/medley/broker.go
ezdac: To the first question: yes, e.g. in time critical applications where it's more important for all receivers to get a chance at receiving than to have delivery guarantees. Also, you can mitigate slow receivers with buffering the channel - they will be able to still get the value as soon as they are idle.
If one receiver still blocks forever on a channel with a full buffer, at least all other channels can still receive.
It is used in the EpochTicker, although this might change since it still has unfavourable properties (as noted in the source).

The initialisation is a good point. We could negate it and call it `nonBlockingSend` if we think this is the default case. One thing to note, a blocking send can be equally bad if one receiver might block forever, so a sensible choice has to be made anyways.

rolling-shutter/medley/broker.go
ezdac: So this one tries a receive on the passed in channel, 
while the case above sends the channel itself to the `b.unsubscribe` channel ;)

It is there because the `Unsubscribe()` e.g. could be called from the same goroutine that usually consumes values
from the channel. But in a blockingSend broker, the broker's loop could be in the process of publishing values, thus it depends on all subscribers still being able to consume the values. The noop `<-channel` allows this while waiting for the channel to be unsubscribed.
Which in turn can only happen if the brokers loop is able to finish the publishing and can receive on the `b.unsubscribe` channel. 
Does this makes sense? I know it's a lot ...

rolling-shutter/medley/broker.go
ezdac: The reason for the initialSubscriptions is the following:

I wanted to be able to have a receive on the subscribed channels waiting in a (or multiple) goroutine and then start the broker. That way it is impossible to miss the first published value.
This would not be easily possible with your second suggestion.

Another way to achieve this would be to make the Subscribe( ) append to the initial subscriptions directly when the broker is stopped, which comes close to your suggestion.
It's possible, but then we would have a `subscriptions` field on the object rather than local state in the Start( ) function.

Currently I favour your first suggestion because this would remove the inconsistency and the slightly confusing default buffer-size. It would also provide more flexibility to the caller of `Subscribe( )`

rolling-shutter/medley/broker.go
ezdac: True, I think this is an automatism of mine, caused by extensive use of `markdown syntax`

rolling-shutter/medley/broker.go
ezdac: Changed the default to `nonBlockingSend=false`

rolling-shutter/medley/broker.go
ezdac: Now the `Subscribe()` takes a channel argument, the buffer size is removed, and subscribe appends to a list of initial subscriptions when the loop is not running.

rolling-shutter/medley/broker.go
ezdac: the buffer is removed now and the callers decide on the channel buffer

PR #319, opened 2022-08-04
WIP include further changes in batchhandler to feature branch

ezdac:


schmir-at-bb:
== Remove State.Process / State.Post? 

The code in `Batch.Run` looks too complicated to me. The different states (we can currently be in define handlers like `OnDecryptionKey`, `OnBatchConfirmation` and so on.  These handlers however return only new states and the real work is done in the `Process` and `Post` functions. We should get rid of the `Process` and `Post` functions and move the code to the handlers, so they do not only decide on what the next state (enum) is, but also how the state changes (i.e. change the batch). 

== Remove/replace OnStateChangePrevious

`Batch.Run` broadcasts state changes here  https://github.com/shutter-network/rolling-shutter/blob/7078d5d9ab4aebfba105ccc52a95ccfb6df72b4f/rolling-shutter/collator/batchhandler/batch/batch.go#L372 and acts on those state changes here https://github.com/shutter-network/rolling-shutter/blob/7078d5d9ab4aebfba105ccc52a95ccfb6df72b4f/rolling-shutter/collator/batchhandler/batch/batch.go#L394-L400

I think we should be able to get rid of `OnStateChangePrevious` and implement the functionality in an existing handler or during the initialization in `Batch.Run`

ezdac:
> == Remove/replace OnStateChangePrevious

The `Publish` channel send you mention will submit the state change of this batch instance to all subscribers (most importantly a `BatchHandler.HandleStateTransitions( )` as well as the `batch.Run( )` of the batch with the "next" batch-index.
The `OnStateChangePrevious` handler will consume an observer channel, which is subscribed to the previous batch,
i.e the handler of batch with index 2 will listen to the state-changes published in the `Run( )` method of the batch with index 1.
So the batch doesn't act on its own state-changes, but on state-changes from the batch before it.

schmir-at-bb:
> So the batch doesn't act on its own state-changes, but on state-changes from the batch before it.

I didn't get that. Thanks for the explanation.

PR #320, opened 2022-08-05
Mock sequencer

### LINKED ISSUE #314: Setup mock sequencer
We need a subcommand for rolling shutter that runs a node exposing a JSON RPC server. The command could be called `mock-sequencer`. It can largely be based on the [mock-eth-server](https://github.com/shutter-network/rolling-shutter/blob/5136ad6cc5db8437a73a29d74ecbcb7f8644a9b7/rolling-shutter/collator/batch/mockethserver.go) we already have for testing the sequencer. Some of the changes that need to be made:

- don't use `testing.T`, but return errors instead
- don't store balances, but return max uint256
- don't store blocks, but return a hardcoded one
- don't store txs
- no need for hooks
- nonces have to be tracked eventually, but for now we can just return `0` for all of them

### LINKED ISSUE #315: Add endpoint to set eon keys to mock sequencer
The mock sequencer should expose an endpoint `admin_addEonKey` that takes an eon key and an activation block number and stores it in its local state. If an eon key for the given activation block number already exists, it should be overridden.

There should be a function to get the eon key for a given block number.

### LINKED ISSUE #316: Add endpoint to set collator to mock sequencer
The mock sequencer should expose an endpoint `admin_addCollator` that takes an address and an activation block number and stores it in its local state. If a collator for the given activation block number already exists, it should be overridden.

There should be a function to get the activation block number for a given block number.

### LINKED ISSUE #317: Add submit batch endpoint to mock sequencer
The endpoint should be called `shutter_submitBatch` and take a serialized batch as parameter. Execution should perform the following steps:

- parse batch
- check chain id
- check and increment batch index
- check decryption key against current eon key (identified by L1 block number)
- check batch signature against current collator (identified by L1 block number)
- decode and decrypt transactions (but don’t execute them)
- check and increment nonces for transactions
- return an error if parsing or any of the above steps fails, otherwise return success

If no collator or eon key is found, an error should be returned.

### LINKED ISSUE #318: Add endpoint to get current batch index to mock sequencer
The endpoint could be called `shutter_getBatchIndex` and should return the current batch index (i.e. the batch index of the last executed batch). If no batch has been executed yet, it should return `-1` or an error.

ezdac:
Fixes #314
Fixes #315 
Fixes #316
Fixes #317
Fixes #318

### DIFF
--- a/rolling-shutter/cmd/mocksequencer/mocksequencer.go
+++ b/rolling-shutter/cmd/mocksequencer/mocksequencer.go
@@ -1,0 +1,58 @@
+package mocksequencer
+
+import (
+	"context"
+	"math/big"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
+	"github.com/spf13/cobra"
+
+	"github.com/shutter-network/rolling-shutter/rolling-shutter/cmd/shversion"
+	"github.com/shutter-network/rolling-shutter/rolling-shutter/mocksequencer"
+)
+
+func Cmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "mock-sequencer",
+		Short: "Run a node that pretends to be a sequencer",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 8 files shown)


jannikluhn (approved):
see comments on branch

PR #321, opened 2022-08-06
Ezdac/collator batch handler

schmir-at-bb:
@ezdac I'm opening this one because I do have some comments on the broker implementation.

### DIFF
--- a/rolling-shutter/medley/broker.go
+++ b/rolling-shutter/medley/broker.go
@@ -1,0 +1,189 @@
+package medley
+
+import (
+	"sync"
+
+	"github.com/rs/zerolog/log"
+)
+
+// Broker allows to distribute a value to multiple
+// receiving channels, so that all channels receive the same value.
+// This is different to the Fan-Out pattern where multiple channels
+// are listening for a send on one channel, but only the first channel
+// to select the receive operation receives this value.
+//
+// The Broker allows to set whether the internal send operation is
+// blocking or non blocking. In the blocking case, all receiving
+// (subscribed) channels have to successively select the receive in order
+// for the broker to continue operation. So if one receiver blocks,
+// everything else will block as well.
+// In the non-blocking case, a subscribed channel that is blocking
+// while the published value is sent will miss the value.
+//
+// One has to be careful in when to select which case.
+// The blocking behavior can be partially subverted by increasing
+// the subscribers buffer size - but if a subscriber is blocked forever,
+// this will only work until this subscribers buffer is full.
+type Broker[T any] struct {
+	mux                  sync.RWMutex
+	stop                 chan struct{}
+	publish              chan T
+	subscribe            chan chan T
+	unsubscribe          chan chan T
+	initialSubscriptions []chan T
+	running              bool
+	nonBlockingSend      bool
+}
+
+func NewBroker[T any](nonBlocking bool) *Broker[T] {
+	return &Broker[T]{
+		stop:                 make(chan struct{}),
+		publish:              make(chan T, 1),
+		subscribe:            make(chan chan T),
+		unsubscribe:          make(chan chan T),
+		nonBlockingSend:      nonBlocking,
+		initialSubscriptions: []chan T{},
+		running:              false,
+	}
+}
+
+// Start starts the Broker's internal loop that processes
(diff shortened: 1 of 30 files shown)


rolling-shutter/medley/broker.go
```diff
@@ -0,0 +1,189 @@
...
+	for _, sub := range b.initialSubscriptions {
+		log.Debug().Int("num-subscribers", len(subscriptions)).Msg("subscribed initial subscription")
+		subscriptions[sub] = struct{}{}
+	}
+	started := make(chan struct{})
+	go func() {
+		b.mux.Lock()
+		b.running = true
+		b.mux.Unlock()
+		close(started)
```
schmir-at-bb, line 69: There's no need for the `started` channel here. Just set the running flag outside of the go routine. Also starting a go routine in the background always looks kind of fishy to me. It might be better to manage that in a error group.

rolling-shutter/medley/broker.go
```diff
@@ -0,0 +1,189 @@
...
+				b.mux.Unlock()
+				return
+			case msgCh := <-b.subscribe:
+				subscriptions[msgCh] = struct{}{}
+			case msgCh := <-b.unsubscribe:
+				delete(subscriptions, msgCh)
+				close(msgCh)
+				log.Debug().Int("num-subscribers", len(subscriptions)).Msg("Unsubscribed")
+			case msg := <-b.publish:
+				if b.nonBlockingSend {
```
schmir-at-bb, line 88: I think the blocking and non blocking cases are mixed up.

rolling-shutter/medley/broker.go
```diff
@@ -0,0 +1,189 @@
...
+		stop:                 make(chan struct{}),
+		publish:              make(chan T, 1),
+		subscribe:            make(chan chan T),
+		unsubscribe:          make(chan chan T),
+		nonBlockingSend:      nonBlocking,
+		initialSubscriptions: []chan T{},
+		running:              false,
+	}
+}
+
```
schmir-at-bb, line 49: The broker code can be much simplified if you replace `NewBroker` and the `Start` method with a function `StartNewBroker` that creates a broker and starts the go routine. You can get rid of the `running` flag, the `mux` field as well as the `initialSubscriptions` field (if there is a need, this can be a argument to `StartNewBroker` and the related logic. There's no need for locking since the struct becomes immutable.

rolling-shutter/medley/broker.go
```diff
@@ -0,0 +1,189 @@
...
+				}
+				b.mux.Lock()
+				b.running = false
+				b.mux.Unlock()
+				return
+			case msgCh := <-b.subscribe:
+				subscriptions[msgCh] = struct{}{}
+			case msgCh := <-b.unsubscribe:
+				delete(subscriptions, msgCh)
+				close(msgCh)
```
schmir-at-bb, line 85: I don't think it's the broker's responsibility to close the `msgCh` here as well as when stopping the broker. We  didn't create the channel, we should not close it.

rolling-shutter/medley/broker.go
ezdac: yes very true - this happened during the broker PR fixups

rolling-shutter/medley/broker.go
ezdac: but isn't the [channel closing principle](https://stackoverflow.com/a/36613932/5053253) about who is sending / receiving values on it rather than who created it?
I think the sender to a channel is supposed to close it, so that on a receive we can check whether there will be more values coming, or whether we can disregard the channel:
```go
newVal, more := <- msgChan
if !more {
   msgChan = nil
}
```

schmir-at-bb: well, the case I'm thinking about is that one could use the same channel to subscribe to multiple brokers. It probably doesn't matter for our use case, so you can also keep the code as is. Though in principle I still do think the we should either manage lifetimes of individual channels completely inside or completely outside the broker.

schmir-at-bb:
This was merged into the `feature/collator-batch-handler`, which itself has been merged into `main` as of 15049dff3f0cb414b5b22a0340c5

PR #322, opened 2022-08-11
Mock-sequencer improvements

ezdac:
### Changes
This PR improves the logging and transaction validation logic of the mock-sequencer.
- activation block based collator retrieval
- activation block based eon key retrieval (although encryption is not yet implemented)
- l1blocknumber matching between shutter-tx's and the parent batch-tx
- pretty logging of shutter-tx and batch-tx
- further batch-tx verification improvements
- bump `txtypes` version to `v0.0.5`

~~#### Missing:
Once https://github.com/shutter-network/go-ethereum/pull/6 is merged and the changes are reflected in the`txtypes` repository,
the imported `txtypes` version  in rolling-shutter has to be updated in this PR.
Otherwise the JSON-marshaling of the transaction logging will not produce the correct transaction structs.~~

### DIFF
--- a/rolling-shutter/mocksequencer/processor.go
+++ b/rolling-shutter/mocksequencer/processor.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
 	txtypes "github.com/shutter-network/txtypes/types"
@@ -133,7 +134,12 @@ func (proc *SequencerProcessor) getNonce(a common.Address, block string) uint64 
 	return nonce
 }
 
-func (proc *SequencerProcessor) processEncryptedTx(txBytes []byte) error {
+func (proc *SequencerProcessor) processEncryptedTx(
+	ctx context.Context,
+	gasPool *core.GasPool,
+	batchIndex, batchL1BlockNumber uint64,
+	txBytes, decryptionKey, eonKey []byte,
+) error {
 	var tx txtypes.Transaction
 	err := tx.UnmarshalBinary(txBytes)
 	if err != nil {
@@ -147,11 +153,29 @@ func (proc *SequencerProcessor) processEncryptedTx(txBytes []byte) error {
 	if err != nil {
 		return errors.New("sender not recoverable")
 	}
+
+	shutterTxStr, _ := tx.MarshalJSON()
+	log.Ctx(ctx).Info().Str("signer", sender.Hex()).RawJSON("transaction", shutterTxStr).Msg("received shutter transaction")
+
+	if tx.L1BlockNumber() != batchL1BlockNumber {
+		return errors.New("l1-block-number mismatch")
+	}
+
+	if tx.BatchIndex() != batchIndex {
+		return errors.New("batch-index mismatch")
+	}
+	_ = decryptionKey
+	_ = eonKey
 	nonce := proc.getNonce(sender, "latest")
 	if tx.Nonce() != nonce+1 {
-		log.Info().Msg("nonce mismatch")
-		return nil
+		return errors.New("nonce mismatch")
 	}
+	err = gasPool.SubGas(tx.Gas())
+	if err != nil {
+		// can return core.ErrGasLimitReached
+		return err
+	}
+
 	proc.setNonce(sender, nonce+1, "latest")
 	return nil
 }
--- a/rolling-shutter/mocksequencer/processor.go
+++ b/rolling-shutter/mocksequencer/processor.go
@@ -2,6 +2,8 @@ package mocksequencer
 
 import (
 	"math/big"
+	"sort"
+	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/pkg/errors"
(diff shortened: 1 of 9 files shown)


rolling-shutter/mocksequencer/processor.go
```diff
@@ -1,24 +1,80 @@
...
+	}
+	if len(blocks) == 0 {
+		return foundVal, errors.New("no value was set")
+	}
+
+	sort.Slice(blocks, func(i, j int) bool { return blocks[i] < blocks[j] })
+
+	// search for index with next highest activation block
+	idx := sort.Search(len(blocks), func(i int) bool {
+		return blocks[i] > block
```
jannikluhn, line 64: why not search for `blocks[i] >= block` directly?

rolling-shutter/mocksequencer/processor.go
```diff
@@ -1,24 +1,80 @@
...
+	})
+
+	// now use the value before that
+	activationBlockIndex := blocks[idx-1]
```
jannikluhn, line 68: it seems `idx` can be zero (e.g. `activation blocks == [5]`, `block == 4`). I guess we should return an error in that case.

rolling-shutter/mocksequencer/processor_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"testing"
+
+	"gotest.tools/assert"
+)
+
+func TestActivationBlockMap(t *testing.T) {
+	mp := newActivationBlockMap[uint64]()
+	mp.Set(uint64(666), 5)
```
jannikluhn, line 11: explicit casting to `uint64` should be unnecessary here

jannikluhn: for readability, using `string` values might be a bit easier though (makes it easier to decide between value and block number)

rolling-shutter/mocksequencer/shutter.go
```diff
@@ -25,7 +27,43 @@ func (s *ShutterService) GetBatchIndex() (string, error) {
...
+	}
+	if collator != sender {
+		return errors.Wrap(err, "not signed by correct collator")
+	}
+
+	// all checks passed, the batch-tx is valid
+	return nil
+}
+
+func (s *ShutterService) SubmitBatch(ctx context.Context, transaction string) (string, error) {
```
jannikluhn, line 66: I think it would be clearer to rename `transaction` to `batchTransaction` (similar to derived variables). Otherwise there's potential to confuse it with transactions included in the batch

rolling-shutter/mocksequencer/shutter.go
```diff
@@ -25,7 +27,43 @@ func (s *ShutterService) GetBatchIndex() (string, error) {
...
+	if tx.Type() != txtypes.BatchTxType {
+		return errors.New("unexpected transaction type")
+	}
+	// TODO retrieve the collator for that block number (which could be higher then the last set
+	// "activation blocknumber")
+	if tx.Gas() > GasLimit {
+		return errors.New("gas limit reached")
+	}
+
+	// tx.Nonce() - ignore the nonce for now
```
jannikluhn, line 40: Batch txs don't have a nonce

rolling-shutter/mocksequencer/shutter.go
```diff
@@ -25,7 +27,43 @@ func (s *ShutterService) GetBatchIndex() (string, error) {
...
 }
 
-func (s *ShutterService) SubmitBatch(transaction string) (string, error) {
+func (s *ShutterService) validateBatch(tx *txtypes.Transaction) error {
+	if tx.Type() != txtypes.BatchTxType {
+		return errors.New("unexpected transaction type")
+	}
+	// TODO retrieve the collator for that block number (which could be higher then the last set
+	// "activation blocknumber")
+	if tx.Gas() > GasLimit {
```
jannikluhn, line 36: Batch txs don't have a gas field (we might want to check the total gas of all transactions though)

rolling-shutter/mocksequencer/shutter.go
```diff
@@ -25,7 +27,43 @@ func (s *ShutterService) GetBatchIndex() (string, error) {
 	return hexutil.EncodeUint64(s.processor.batchIndex), nil
 }
 
-func (s *ShutterService) SubmitBatch(transaction string) (string, error) {
+func (s *ShutterService) validateBatch(tx *txtypes.Transaction) error {
+	if tx.Type() != txtypes.BatchTxType {
+		return errors.New("unexpected transaction type")
+	}
+	// TODO retrieve the collator for that block number (which could be higher then the last set
```
jannikluhn, line 34: comment can be removed, it's done below

rolling-shutter/mocksequencer/shutter.go
```diff
@@ -25,7 +27,43 @@ func (s *ShutterService) GetBatchIndex() (string, error) {
...
+	}
+	sender, err := s.processor.signer.Sender(tx)
+	if err != nil {
+		return errors.Wrap(err, "error recovering batch tx sender")
+	}
+	if collator != sender {
+		return errors.Wrap(err, "not signed by correct collator")
+	}
+
+	// all checks passed, the batch-tx is valid
```
jannikluhn, line 62:
```suggestion
	// all checks passed, the batch-tx is valid (disregarding validity of included transactions)
```

rolling-shutter/mocksequencer/shutter.go
```diff
@@ -36,24 +74,41 @@ func (s *ShutterService) SubmitBatch(transaction string) (string, error) {
...
+		log.Ctx(ctx).Error().Err(err).Msg("error while retrieving eon key")
+		return "", err
+	}
+
 	for _, shutterTx := range tx.Transactions() {
-		err := s.processor.processEncryptedTx(shutterTx)
+		err := s.processor.processEncryptedTx(ctx, tx.BatchIndex(), tx.L1BlockNumber(), shutterTx, tx.DecryptionKey(), eonKey)
 		if err != nil {
-			log.Error().Err(err).Msg("tx not processable, dropping transaction")
+			log.Error().Err(err).Msg("transaction invalid")
```
jannikluhn, line 101: Invalid transactions should result in the batch being invalid as well (at least for everything that the collator can control, i.e. all plaintext fields -- but we're not checking the decrypted part yet anyways).

jannikluhn (approved):

rolling-shutter/mocksequencer/processor.go
```diff
@@ -1,24 +1,80 @@
...
+)
+
 type blockData struct {
 	baseFee  *big.Int
 	gasLimit uint64
 }
 
+type activationBlockMap[T any] struct {
+	mux sync.RWMutex
+	mp  map[uint64]T
```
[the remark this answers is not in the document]
ezdac: i think efficiency is not relevant here, especially since the list will be small in testing cases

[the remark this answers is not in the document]
jannikluhn: agreed

rolling-shutter/mocksequencer/processor.go
ezdac: we actually are looking for `i` such that`blocks[i] <= block` AND  `blocks[i+1] > block` but only if there is a higher block. Searching for `blocks[i] <= block` was not possible because of some properties of the search function I believe.

I decided to not try to encode everything in the closure (also because of index out of bounds handling) but rather search the first block that doesn't match the desired condition anymore. The search returns `len(blocks)` when no such index is present.

ezdac: But I found another solution that simplifies the code

rolling-shutter/mocksequencer/processor_test.go
ezdac: Not sure what you mean - make the map a "string" valued map?
I will not implement this for now because I think it's not that important and I don't want to deal with the type-conversion when using that map and comparing against tx-fields (uint64)

jannikluhn: Sorry, I mean just for this test. Activation block maps are generic, so you could use `string` as the type parameter and store `"first entry"` starting at block 0, `"second entry"` starting at block 5, etc.

ezdac: oh ok that makes sense

PR #330, opened 2022-10-11
Add batcher package

schmir-at-bb:
This is an implementation of the batchhandler functionality in a synchronous way. It doesn't start multiple go routines and should be easier to maintain.


ezdac:
LGTM

PR #338, opened 2022-11-09
Remove batchHandler field from collator struct

schmir-at-bb:
This removes the use of the batch handler from the collator and replaces it with a batcher instance.

We still have a lot of out-commented code that can be either adapted or removed later.

### DIFF
--- a/rolling-shutter/collator/http.go
+++ b/rolling-shutter/collator/http.go
@@ -69,13 +69,7 @@ func (srv *server) SubmitTransaction(w http.ResponseWriter, r *http.Request) {
 	// during inclusion in the batch, e.g. because of nonce mismatch
 	// or lack of funds
 
-	// for now this waits until the final async result is set on the tx.
-	// this will block the request potentially for
-	// quite some time (until there is finality on the tx),
-	// worst case until the tx has been included in the batch and
-	// the batch has been confirmed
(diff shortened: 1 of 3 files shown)


schmir-at-bb:
If this causes too much merge conflicts on your side, we can also defer merging this PR

rolling-shutter/collator/http.go
```diff
@@ -69,13 +69,7 @@ func (srv *server) SubmitTransaction(w http.ResponseWriter, r *http.Request) {
...
 	// or lack of funds
 
-	// for now this waits until the final async result is set on the tx.
-	// this will block the request potentially for
-	// quite some time (until there is finality on the tx),
-	// worst case until the tx has been included in the batch and
-	// the batch has been confirmed
-	res := <-srv.c.batchHandler.EnqueueTx(ctx, x.EncryptedTx)
-	err := res.Err
+	err := srv.c.batcher.EnqueueTx(ctx, x.EncryptedTx)
```
ezdac, line 72: I don't think it's necessary for this PR, but we need to think about concurrency of this method call.
Worst case would be to serialise the call here.

ezdac (approved):
Looks good - but I would prefer to remove the commented-out code.

rolling-shutter/collator/http.go
ezdac: Ah yes, I didn't catch that

Issue #332, opened 2022-11-03
Collator: send DecryptionTrigger to P2P-messaging


ezdac:
After we merged #330, the integration to the messaging layer still needs to be implemented.
The `batcher` puts a DecryptionTrigger in the collator-DB after the batch was closed via the `batcher.CloseBatch( )` method.
A dedicated routine needs to:
- read new, unsent decryption-triggers from the DB
- encode and sign a `shmsg.DecryptionTrigger` from that data
- forward it to the P2P messaging layer
- mark the DecryptionTrigger as sent in the DB

ezdac:
closing, since #339 is merged

PR #339, opened 2022-11-21
Send decryption-trigger to P2P messaging

ezdac:
implements #332 

https://github.com/shutter-network/rolling-shutter/pull/339/commits/fd958e77fe2afd0760f2e8ff6f7f84fe51032429 additionally refactors the `medley.Retry` function-call retry to take additional, but optional parameters regarding logging and retry-metrics like time-interval, exponential-backoff and number of retries.


schmir-at-bb (approved):
nice!

PR #340, opened 2022-12-08
Collator: batch-closing routine

### LINKED ISSUE #335: Collator: close batch every X seconds
In order to progress shutter-batches on the rollup, the collator needs to trigger the commitment and decryption of the current batch being constructed. Therefore a parallel process has to close current batches (`batcher.CloseBatch( )`) as soon as the following conditions are met:
- the previous batch / epoch has to be confirmed and included in the rollup (poll the `shutter_batchNumber` JSON RPC endpoint)
- the time passed since the last batch has been closed has to be at least `X` seconds

A fitting value for parameter `X` has to be empirically derived, depending on the targeted rollup and application context.

ezdac:
closes #335 

### What is implemented in this PR?
1) A new routine is added to the collator which closes batches at a constant interval by calling `batcher.CloseBatch()` and afterwards trying to initiate the batchers chain-state by spamming `batcher.EnsureChainState()`
2) `medley.retry` now has the `StopOnError` and `NumberOfRetries(-1)` configuration options
3) a fix in the `retry.retrier.iterator` implementation + minor changes in the containing package
4) expose error instances from `batcher` publicly

### DIFF
--- a/rolling-shutter/collator/batcher/batcher.go
+++ b/rolling-shutter/collator/batcher/batcher.go
@@ -95,6 +95,15 @@ func (btchr *Batcher) earlyValidateTx(tx *txtypes.Transaction) error {
 	return nil
 }
 
+func (btchr *Batcher) EnsureChainState(ctx context.Context) error {
+	btchr.mux.Lock()
+	defer btchr.mux.Unlock()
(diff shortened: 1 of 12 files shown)


schmir-at-bb (approved):
LGTM

Issue #347, opened 2023-01-11
Progress batches only when a valid Eon configuration is known


ezdac:
To be checked before progressing batches in the batcher:
- Eon public key
- valid Keyper set

In order to get that information, the `L1 block-number` associated with that batch is required.

PR #350, opened 2023-01-16
Do not send close batch if we have no eon public key

schmir-at-bb:
We lookup the eon public key before closing a batch and reject doing that if we don't have an eon public key.

See #347

### DIFF
--- a/rolling-shutter/collator/batcher/batcher.go
+++ b/rolling-shutter/collator/batcher/batcher.go
@@ -232,6 +232,13 @@ func (btchr *Batcher) closeBatchImpl(ctx context.Context, db *cltrdb.Queries, l1
 		return err
 	}
 
+	// Lookup the current eon public key for the given block.
+	_, err = db.FindEonPublicKeyForBlock(ctx, l1blockNumber)
+	if err != nil {
+		log.Info().Int64("l1BlockNumber", l1blockNumber).Msg("no eon public key found")
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


schmir-at-bb:
Looks like I also need to adapt the tests...

schmir-at-bb:
@ezdac fixed the tests. please take a look at it

rolling-shutter/collator/batcher/batcher.go
```diff
@@ -232,6 +232,13 @@ func (btchr *Batcher) closeBatchImpl(ctx context.Context, db *cltrdb.Queries, l1
 		return err
 	}
 
+	// Lookup the current eon public key for the given block.
+	_, err = db.FindEonPublicKeyForBlock(ctx, l1blockNumber)
+	if err != nil {
```
ezdac, line 237: nitpick: this could also be a database failure or any other kind of error apart from `pgx.ErrNoRows`

Returning different errors for different conditions wouldn't change anything in the calling code though, since we don't catch specific errors there.

ezdac (approved):
Looks good! 

In order to close #347 we would need to add a check for an
existing keyper set in the DB as well.

rolling-shutter/collator/batcher/batcher.go
schmir-at-bb: Right, I think I am a bit lazy with the error handling in general.

PR #358, opened 2023-02-17
Upgrade `libp2p-core` imports to monorepo package imports

ezdac:
The libp2p-core library is being deprecated and is now placed in the main libp2p monorepo.
This also caused some issues updating to golang 1.10 because auf a transient libp2p-core library (IIRC).


schmir-at-bb:
thanks

PR #360, opened 2023-03-03
PeerExchange (PX) and DHT gossipsub discovery

ezdac:
## Peer and topic discovery
Peers in our libp2p messaging network will use [gossipsub 1.1 PX](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#prune-backoff-and-peer-exchange) in order to being bootstrapped for subscribed topics:

A bootstrap node that has PX activated will automatically PRUNE a gossipsub connection of a peer immediatly,
but will provide the pruned node with additional metadata of other peers interested in the topic.
The scoring system has to be adjusted in a certain way to allow peer-exchange, as stated in the [gossipsub 1.1 spec](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#recommendations-for-network-operators).

As an "ambient peer discovery" fallback for routing (finding correct dial endpoints with full multiaddr address paths) 
and topic discovery (finding peers to form the gossipsub overlay mesh for the same topics), a DHT based solution is used. 
However it is not tested if this properly works with the used parameters and configuration.

## Breaking changes
- Fix / change automatic binding of environment variables
- Change configuration options

## Other changes
- Add `p2pnode` command for p2p bootstrap node
- Add bootstrap node integration in clojure code  / e2e-tests


schmir-at-bb:
LGTM. `ROLLING_SHUTTER_ROOT` should not be renamed. We also use the same environment variable in `play.clj`
The let binding is `bb.edn` is hard to read. Values should start on the same line as the variable.

PR #361, opened 2023-03-03
Fix batcher's next-batch always initialized to "1"

ezdac:
Before, the next batch in the database was always initialized to 1 on startup. This was also the case when the sequencer already was at a higher batch number.
This caused a non-recovering error-condition in the `initChainState` method, which was masked in the tests by manually setting the  db's next-batch value during fixture creation.

Now, the collator queries the batch-number of the sequencer on startup, as long as the collator does not have a next-batch set in the DB already.


schmir-at-bb (approved):
LGTM

PR #362, opened 2023-03-14
Split off p2p messages from shmsg

schmir-at-bb:
We defined all of the protoc generated code inside one package. This commit moves the p2p related messages to a dedicated package.


rolling-shutter/p2pmsg/messages.go
```diff
@@ -0,0 +1,180 @@
...
+		if val != messageTypeName {
+			err := errors.Errorf("Topic '%s' already has message type <%s> registered. Registering %s failed", topic, val, messageTypeName)
+			panic(err)
+		}
+	}
+	topicToProtoName[topic] = messageTypeName
+}
+
+// P2PMessage can be send via the p2p protocol.
+type P2PMessage interface {
```
ezdac, line 52: we could rename this now to `p2pmsg.Message`

ezdac (approved):
looks good

rolling-shutter/p2pmsg/messages.go
schmir-at-bb: I'll probably do this in another commit

PR #363, opened 2023-03-22
Untangle keyper pkg

schmir-at-bb:
Split the keyper package in subpackages. This should make reuse of the code easier.

This is not done yet, I also like to create a package for the smdriver/smstate code. But I think we could already merge this one.


rolling-shutter/keyper/kprapi/kprapi.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func (srv *server) Start(ctx context.Context, group *errgroup.Group) error {
+	httpServer := &http.Server{
+		Addr:              srv.config.GetHTTPListenAddress(),
+		Handler:           srv.setupRouter(),
+		ReadHeaderTimeout: 5 * time.Second,
+	}
+	group.Go(httpServer.ListenAndServe)
+	group.Go(func() error {
```
ezdac, line 95: Here is the graceful shutdown pattern again. I use it 2 times in the new tracing / metrics as well. Would make sense to  support this in the medley/service

rolling-shutter/keyper/epochkghandler/epochkghandler.go
```diff
@@ -11,27 +13,50 @@ import (
...
 )
 
-type epochKGHandler struct {
-	config Config
-	db     *kprdb.Queries
+type EpochKGHandler struct {
+	address    common.Address
+	instanceID uint64
+	db         *kprdb.Queries
+	dbpool     *pgxpool.Pool
```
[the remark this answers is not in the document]
schmir-at-bb: no, the whole thing probably needs some more polish. Also the handling of the config object is inconsistent. I'll fix that later.

rolling-shutter/keyper/kprapi/kprapi.go
schmir-at-bb: I thought about having a start/stop method as part of the service interface and wrote some code for it. But it's not straight forward. E.g. in the code above the `httpServer.ListenAndServe` will never return. It' probably best to set the BaseContext field in the http server struct in order to allow shutting down the server by cancelling the context.

PR #366, opened 2023-03-28
Add instrumentation for cross-boundary tracing

ezdac:
Adds all of the necessary setup and instrumentation functions to allow for using Opentelemetry for tracing and metrics
in non-production environments.

This does not include:
- manual instrumentation across the codebase (but this requires to only add 2 LoC per span)
- setup and enabling of the tracing for the command entrypoints
- configuration option to enable tracing per command


schmir-at-bb (approved):
lgtm

PR #367, opened 2023-04-27
Fix bug in keyper config

konradkonrad:
The config key is `ListenAdresses` (plural).


konradkonrad:
@ezdac

ezdac:
This ist not relevant anymore since we have the refactored config system.

PR #368, opened 2023-05-03
Snapshot/merge

konradkonrad:
This PR is a "port" of the snapshot specific code that lived in branch `snapshot/main` onto the current `main` branch. The intention is to synchronize with the tooling/patterns used in `main` and benefit from newer features, e.g. p2p discovery.
For the future, this should allow to share the common code base and concentrate on the snapshot shutter specific bits.


contracts/deploy/515_configure_collator.js
```diff
@@ -36,17 +37,18 @@ module.exports = async function (hre) {
...
   const activationBlockNumber = currentBlock + 10;
-
   const activeConfig = await cfg.getActiveConfig(activationBlockNumber);
   if (activeConfig[1].toNumber() === configSetIndex) {
     console.log("Collator config already added");
     return;
   }
 
-  await cfg.addNewCfg({
+  configSetIndex--;
```
konradkonrad, line 46: @ulope this "hack" seems to do the trick in the clean slate deployment case, but it's not correct

rolling-shutter/snapshot/handler.go
```diff
@@ -0,0 +1,154 @@
...
+	return []p2pmsg.Message{&p2pmsg.DecryptionTrigger{}}
+}
+
+func (*EonPublicKeyHandler) MessagePrototypes() []p2pmsg.Message {
+	return []p2pmsg.Message{&p2pmsg.DecryptionTrigger{}}
+}
+
+func (handler *DecryptionKeyHandler) ValidateMessage(ctx context.Context, msg p2pmsg.Message) (bool, error) {
+	decryptionKeyMsg := msg.(*p2pmsg.DecryptionKey)
+	//FIXME: check snapshot business logic for decryptionKeyMsg validation
```
konradkonrad, line 57: @ulope I don't know if there is more to do here?

rolling-shutter/snapshot/handler.go
```diff
@@ -0,0 +1,154 @@
+package snapshot
```
konradkonrad, line 1: @ulope I created this whole message handler module to mirror the other message handling code in the main branch. I have _not_ run any tests to confirm, that it actually works as intended.

rolling-shutter/snapshot/handler.go
```diff
@@ -0,0 +1,154 @@
...
+		return nil, err
+	}
+	seenEons[eonId] = struct{}{}
+
+	return nil, nil
+}
+
+func (handler *TimedEpochHandler) HandleMessage(ctx context.Context, m p2pmsg.Message) ([]p2pmsg.Message, error) {
+	var result []p2pmsg.Message
+	//FIXME: add TimedEpoch handling logic
```
konradkonrad, line 152: @ulope This one is still not implemented

PR #369, opened 2023-05-03
Run make-generate

konradkonrad:
I get different ABIs when I run `make-generate` from current `main` branch (https://github.com/shutter-network/rolling-shutter/commit/ae8efa860ec800da61e5941d2abc6b67b7549012)


schmir:
I've changed the CircleCI settings to build pull requests, which it didn't do and also forced a build of this pull request.
As you can see, the `rs-generate` step fails. Not sure what's wrong at the moment. Let's discuss in the daily.

PR #373, opened 2023-06-01
Use crypt/rand for Message nonce

konradkonrad:
This fixes a linting error by `gosec` - using `math/rand` is discouraged.


konradkonrad:
@jannikluhn I'm not entirely sure if a `random` nonce is the right thing to do here. Can you review this?

PR #374, opened 2023-06-05
Add generate to pre-commit

ulope:
Add a local hook to run `make generate` in pre-commit to prevent commits with missing or outdated generated files.


.pre-commit-config.yaml
```diff
@@ -2,6 +2,16 @@ default_language_version:
   python: python3
 
 repos:
+  - repo: local
+    hooks:
+      - id: generate
+        name: generate
+        language: system
+        entry: bash -c 'cd rolling-shutter && make generate'
+        always_run: true
```
konradkonrad, line 11: why do we have to `always_run: true` here?
couldn't we also use some rather broad [type filters](https://pre-commit.com/#filtering-files-with-types) instead?

ulope: Yeah probably, I started with writing a `files:` filter but then abandoned that idea since I didn't want to spend too much time on it and also it seemed like one of those things that get forgotten to update when at some point a new `// go:generate` directive is added somewhere.

With already existing files `make generate` takes less than 3s on my machine. That seemed acceptable.

PR #381, opened 2023-06-22
Setup distributed Shuttermint in snapshutter

### LINKED ISSUE #380: Set up distributed shuttermint for snapshutter
The docker-compose setup currently runs a single shuttermint node in `dev` mode.

In production, each keyper should run their own shuttermint chain node alongside.

konradkonrad:
Notable changes included:
- Write the validator public key of a TM node into it's own file
- Add `bb` test for changing the initial validator set
- Run 3 TM `chain`s
- removed `--dev` mode from TM command

Fixes #380


schmir:
This breaks the principle of least privilege.

konradkonrad:
> This breaks the principle of least privilege.

This comment referred to a change that read the private key. This is no longer the case. Thanks schmir!

PR #387, opened 2023-07-17
Allow specifying the initial eon when initializing the chain

schmir:
This allows specifying the initial eon when initializing a chain.


ulope (approved):
LGTM

Issue #388, opened 2023-07-17
segfault when shutting down chain subcommand


schmir:
```
$ rm -rf boot; rolling-shutter chain init --root boot --genesis-keyper 0x440Dc6F164e9241F04d282215ceF2780cd0B755e; rolling-shutter chain --config boot/config/config.toml --loglevel=debug
2023/07/17 17:06:50.322511 INF [         init.go:203] Generated private validator privValKeyFile=boot/config/priv_validator_key.json stateFile=boot/data/priv_validator_state.json
2023/07/17 17:06:50.322613 INF [         init.go:212] Saved private validator publickey path=boot/config/priv_validator_pubkey.hex validatorPublicKey=e1513d8d803bafd7f69bb4633814b4f8f428e818e417a1d605c18d4908505471
2023/07/17 17:06:50.322706 INF [         init.go:242] Generated genesis file path=boot/config/genesis.json
2023/07/17 17:06:50.322765 INF [         init.go:259] Generated node key id=de4c26ba15043483d46cfad78f9eff71fd4c35f3 path=boot/config/node_key.json
2023/07/17 17:06:50.322781 INF [          app.go:753] persisting state to disk height=0
2023/07/17 17:06:50.355138 INF [         chain.go:40] starting shuttermint version="(devel-075ecfa5e8776562a266baacceb73bbdc5288bb8) (go1.20.3, linux-amd64)"
2023/07/17 17:06:50.356174 INF [         chain.go:68] loaded node-id node-id=de4c26ba15043483d46cfad78f9eff71fd4c35f3
2023/07/17 17:06:50.356374 INF [          app.go:111] Loaded shutter app from file devmode=false file=boot/data/shutter.gob last-block-height=0 last-saved="2023/07/17 15:06:50.322810"
2023/07/17 17:06:50.451456 INF [          app.go:230] initializing new chain
2023/07/17 17:06:50.451475 INF [          app.go:232] initial keyper index=0 keyper="0x440Dc6F164e9241F04d282215ceF2780cd0B755e [chksum ok]"
^C2023/07/17 17:06:51.588071 INF [       service.go:84] received OS signal, shutting down signal=interrupt
2023/07/17 17:06:51.588180 DBG [         chain.go:97] Stopping node
2023/07/17 17:06:51.588275 ERR [         node.go:820] failed to stop event sink err="leveldb: closed"
panic: runtime error: invalid memory address or nil pointer dereference
[...]

PR #389, opened 2023-07-20
Upgrade tendermint to v0.37.0-rc2

schmir:
This fixes a segfault when shutting down a tendermint validator. Please note that the tendermint configuration file now mostyly uses snake case instead of kebab case.

See https://github.com/shutter-network/rolling-shutter/issues/388

I'm not completely sure if we should merge this now, since it's a release candidate. On the other hand running without the pex reactor also doesn't seem to be well tested.


schmir:
The docker stuff in the docker directory currently doesn't seem to work, looking into it.

PR #398, opened 2023-08-28
Update go-libp2p to v0.30.0 and go to 1.21

schmir:
Update go-libp2p to v0.30.0 and go to 1.21

This version is compatible with go 1.21 and allows us to build binaries with this go version.


schmir:
Added two more commits: 5fdeef8ca6155dd9a53809c184c2a981f97633cd, 2defe495f250227c10e7a597a6e72f6399305a

PR #404, opened 2023-09-05
Chainobserver: sync from earliest contract deployment only

ezdac:
Before we blindly synced blockchain events from the block 0 on without considering the contract creations.
Now, we take the minimum of all contract creation block-numbers for events we are interested in and start syncing there.
If we saved a sync progress later than that, this is used instead as the starting point for syncing.


ulope (approved):
Lgtm

PR #406, opened 2023-09-07
fix(snapshot): get seen eons / epochs from db instead of map

ezdac:
Before we kept an in-memory map of seen eons and epoch, in order to avoid firing events multiple times.
If the node would crash, this information would be lost - although this has no negative effect on database records.

Now, we simply try to insert incoming objects into our db, and will react on wether `ON CONFLICT DO NOTHING` happened during the insert.


ulope (approved):
LGTM

I also added that already known keys will be served back to snapshot from the DB instead of re-sending a decryptionTrigger.

PR #408, opened 2023-09-11
Speedup 'make generate' and the generate pre-commit hook

schmir:


ulope (approved):
Thank you *very much*

Inspired by this I added a make based speedup for the `gendocs` part as well.

PR #409, opened 2023-09-18
Redact sensitive values on config logging

ezdac:
This introduces another `shconfig` tag `sensitive`.
When using the `LogConfig` function, all values tagged with sensitive will be redacted upon logging and a `(sensitive)` value will be shown instead.
The tree-traversal based dict-parsing function might not be the most efficient, but logging of the config values is expected to be done infrequently / once only.

### DIFF
--- a/rolling-shutter/medley/configuration/traverse.go
+++ b/rolling-shutter/medley/configuration/traverse.go
@@ -1,6 +1,7 @@
 package configuration
 
 import (
+	"encoding"
 	"fmt"
 	"io"
 	"reflect"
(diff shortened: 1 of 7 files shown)


rolling-shutter/medley/configuration/traverse.go
```diff
@@ -38,12 +40,31 @@ func extractTags(field *reflect.StructField) tagOptions {
...
+	_, sensitive := tags["sensitive"]
 	return tagOptions{
 		setExplicitly: true,
 		required:      required,
+		sensitive:     sensitive,
 	}
 }
 
+// GetSensitiveRecursive will return all sensitive config paths.
+func GetSensitiveRecursive(root Config) map[string]any {
```
schmir, line 52: We use `struct{}` as the value type for maps used like sets.  I think we should also do this here as it makes the code a bit clearer. In case I'm  wrong and we need the value, I don't understand why this isn't a bool.

PR #410, opened 2023-09-25
Feature/remove docker sentry nodes

konradkonrad:
This changes the local docker test setup for snapshot shutter to run without `sentry` nodes / without validator isolation.

It changes the `chain init` parameter `--role validator` to use parameters for an exposed validator. The old functionality is preserved in `--role isolated-validator`.

This now is a better match for the keyper configuration in https://github.com/rolling-shutter/snapshot-keyper

### DIFF
--- a/docker/02-init-chain.sh
+++ b/docker/02-init-chain.sh
@@ -16,13 +16,19 @@ ${BB} rm -rf data/deployments
 # has geth as dependency
 $DC up deploy-contracts
 
+# setup chain-seed
 $DC run --rm --no-deps chain-seed init \
     --root /chain \
     --blocktime 1 \
     --listen-address tcp://0.0.0.0:${TM_RPC_PORT} \
     --role seed
 
-for num in 0 1 2 3; do
+seed_node=$(cat data/chain-seed/config/node_key.json.id)@chain-seed:${TM_P2P_PORT}
+
+${BB} sed -i "/^moniker/c\moniker = \"chain-seed\"" data/chain-seed/config/config.toml
+
+# configure validators and keypers 0-3
+for num in {0..3}; do
     validator_cmd=chain-$num-validator
 
     $DC run --rm --no-deps ${validator_cmd} init \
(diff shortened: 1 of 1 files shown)


ulope (approved):
👍

rolling-shutter/cmd/chain/init.go
```diff
@@ -170,24 +180,25 @@ func adjustPort(address string, keyperIndex int) (string, error) {
...
-			Msg("Found private validator")
-	} else {
-		pv = privval.GenFilePV(privValKeyFile, privValStateFile)
-		pv.Save()
-		log.Info().
-			Str("privValKeyFile", privValKeyFile).
-			Str("stateFile", privValStateFile).
-			Msg("Generated private validator")
-	}
+	if slices.Contains([]string{VALIDATOR, ISOLATEDVALIDATOR}, config.Role) {
```
ezdac, line 183: SEED role also needs privkey files

docker/02-init-chain.sh
```diff
@@ -4,71 +4,61 @@ source ./common.sh
...
+    ${BB} sed -i "/ValidatorPublicKey/c\ValidatorPublicKey = \"$(cat data/${validator_cmd}/config/priv_validator_pubkey.hex)\"" /config/keyper-${num}.toml
 
-    # set seed node for sentry
-    ${BB} sed -i "/^persistent_peers =/c\persistent_peers = \"${seed_node}\"" data/${sentry_cmd}/config/config.toml
-    # set validator node for sentry
-    ${BB} sed -i "/^private_peer_ids =/c\private_peer_ids = \"${validator_id}\"" data/${sentry_cmd}/config/config.toml
-    ${BB} sed -i "/^unconditional_peer_ids =/c\unconditional_peer_ids = \"${validator_id}\"" data/${sentry_cmd}/config/config.toml
-    ${BB} sed -i "/^external_address =/c\external_address = \"${sentry_cmd}:${TM_P2P_PORT}\"" data/${sentry_cmd}/config/config.toml
+    # set seed node for chain bootstrap
+    ${BB} sed -i "/^bootstrap_peers =/c\bootstrap_peers = \"${seed_node}\"" "${validator_config_path}"
```
ezdac, line 56: In the 0.0.37-rc2 this is call `seeds` now

ezdac (commented):
Just putting some comments here for reference if needed.
I am currently fixing some of the bugs / syntax errors in a local branch

rolling-shutter/cmd/chain/init.go
ulope: Ehh, are you sure about that?

AFAICS the point of the seed role is to not be an actual part of the network but just to exchange peer information.

ezdac: No not sure, but IIRC the seed role node errored on me because it was looking for the file that was configured in the config.
Before this PR, the seed validator *did* get a privkey file.

So maybe it doesn't need it, but then this needs to be reflected in the tendermint config file as well.
Maybe the config file has to be the empty string then

ezdac: And apart from that - wouldn't it make sense that also seed nodes are authenticated by address /signature?

ulope: They do need a node key (`node_key_file` in the config) for participating on the p2p network but what we're talking about here is the validator key.

I've just checked the boot nodes on the shapshot installation. They definitely don't have the `priv_*` key files.

PR #420, opened 2023-11-14
Database Refactoring

ezdac:
Before, the database schemas were defined in one directory and code generation was configured with one `sqlc.yaml` file. The database schema validation was only conducted against the most specific schema definition for the command that was invoked (e.g. the "collator" schema). This ignored schema changes for definitions that were used across commands, like "chainobsdb" or the keyper database definition.

Now the code generation is split up in multiple sqlc modules that have the same folder structure.
The location of the sql and sqlc files for each individual database module lies within the go package that uses the database the most.
Additionally, database definitions now comply to an interface `Definition` and they are created at initialisation time.
The definitions are used to initialise and validate the database. An implementation `AggregateDefinition` can be used to wrap multiple database module definitions in one `Definition` and thus combine  (hierarchical) definitions into one flat set of database modules. This allows to check for the schema versions of all database modules.


play/process-compose.yaml
```diff
@@ -0,0 +1,26 @@
+version: "0.5"
```
ulope, line 1: Cool, didn't know about process-compose before.

ulope (approved):
Nice, I especially like all the removed duplicated boilerplate!
I can't say that I read every line, but it looks good.

About the CI error, we could increase the resource class of that runner and see if that fixes the issue.

play/process-compose.yaml
ezdac: It's nice, although I had some issues with `depends_on` chains not working properly. But the checked in code here does not have this.

ezdac:
CI memory increase to `medium+` helped.

PR #425, opened 2024-01-10
Support multiple keys in single message

jannikluhn:
With this PR, multiple keys and key shares in a single message are handled correctly. When the keyper receives a decryption trigger with at least one new identity, it generates a decryption key shares message for all identities. When it receives a decryption key shares message that enables it to generate at least one new decryption key, it generates a decryption keys message for all identities. When it receives a decryption keys message, it inserts all keys into the db.

An alternative would be to only consider new identities, eg if the keyper receives a trigger for identities 1 and 2, but for identity 1 the share has already been sent, send a shares message only for identity 2. However, this would be inconvenient at the receiver's side, as effectively a single trigger would be broken up into multiple decryption keys messages, which they would have to piece together in the end. During normal operation (ie all keypers broadcast all requested shares), the outcome is the same.


rolling-shutter/snapshot/handler.go
```diff
@@ -90,13 +79,18 @@ func (handler *DecryptionKeyHandler) ValidateMessage(ctx context.Context, msg p2
...
 	}
 
-	epochID := decryptionKeyMsg.GetEpochID()
-	ok, err := shcrypto.VerifyEpochSecretKey(key, &eonPublicKey, epochID)
-	if err != nil {
-		return pubsub.ValidationReject, err
-	}
-	if !ok {
-		return pubsub.ValidationReject, errors.Errorf("recovery of epoch secret key failed for epoch %s", epochID)
+	for _, key := range keys.Keys {
```
ezdac, line 82: Same question as in the other validation functions

rolling-shutter/snapshot/handler.go
```diff
@@ -117,31 +111,31 @@ func (handler *EonPublicKeyHandler) ValidateMessage(_ context.Context, msg p2pms
...
-	)
-	if err != nil {
-		return result, err
-	}
-
-	// already seen
-	if rows == 0 {
-		return result, nil
+	for _, key := range keys.Keys {
+		rows, err := db.InsertDecryptionKey(
```
ezdac, line 118: here we potentially should first insert all keys and then submit them to the api?
I'm not sure how the recovery case is handled, but I think we don't have retries currently.
So like this at least the keys itself are not lost if there is an arbitrary error in the api.

rolling-shutter/collator/epochhandling.go
```diff
@@ -101,17 +105,24 @@ func (handler *decryptionKeyHandler) ValidateMessage(
...
 
-	ok, err := shcrypto.VerifyEpochSecretKey(epochSecretKey, &eonPublicKey, identityPreimage.Bytes())
-	if err != nil {
-		return pubsub.ValidationReject, err
+	if len(keys.Keys) == 0 {
+		return pubsub.ValidationReject, errors.Errorf("no keys in message")
 	}
-	if !ok {
-		return pubsub.ValidationReject, errors.Errorf("recovery of epoch secret key failed for epoch %s", identityPreimage)
+	for _, key := range keys.Keys {
```
ezdac, line 112: Same as other validation functions

ezdac (changes requested):
Looks good!
I have some small change suggestions and then a DoS-related question concerning all the p2p validation functions.

PR #428, opened 2024-01-22
feat: add gnosiskeyper command

jannikluhn:
Sets up a command to run a gnosis keyper. For now, it doesn't do much beyond running the core keyper.

### DIFF
--- a/rolling-shutter/keyper/kprapi/http.go
+++ b/rolling-shutter/keyper/kprapi/http.go
@@ -32,7 +32,7 @@ func (srv *Server) Ping(w http.ResponseWriter, _ *http.Request) {
 	_, _ = w.Write([]byte("pong"))
 }
 
-func (srv *Server) Shutdown(w http.ResponseWriter, _ *http.Request) {
+func (srv *Server) Shutdown(_ http.ResponseWriter, _ *http.Request) {
 	srv.shutdownSig <- struct{}{}
 	// We still want to return here and thus return 200 to the caller after this.
 	// Not immediately closing open connctions is taken care


rolling-shutter/cmd/gnosiskeyper/gnosiskeyper.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/shutter-network/rolling-shutter/rolling-shutter/keyperimpl/gnosis/database"
+	"github.com/shutter-network/rolling-shutter/rolling-shutter/medley/configuration/command"
+	"github.com/shutter-network/rolling-shutter/rolling-shutter/medley/db"
+	"github.com/shutter-network/rolling-shutter/rolling-shutter/medley/service"
+)
+
+func Cmd() *cobra.Command {
+	builder := command.Build(
+		main,
+		command.CommandName("gnosiskeyper"),
```
ezdac, line 22: I would remove this and use the `Configuration.Name()` function to return the command name.
(See other comment)

rolling-shutter/keyperimpl/gnosis/config.go
```diff
@@ -0,0 +1,72 @@
...
+	Shuttermint *kprconfig.ShuttermintConfig
+	Metrics     *metricsserver.MetricsConfig
+}
+
+func (c *Config) Validate() error {
+	return nil
+}
+
+func (c *Config) Name() string {
+	return "keyper"
```
ezdac, line 47: Here you can change the name to represent the CLI command name you want.
So e.g. `return "gnosiskeyper"` in your case. 
This is a little bit different than chosing the `CommandName` option above and sticks to the convention of other keyperimpl's

ezdac (changes requested):
Looks good so far.
Just needs to change the place where the CLI subcommand name is defined and making the linter happy.

PR #429, opened 2024-01-31
Initial op-shutter keyper

ezdac:


ezdac:
The tests are failing due to (probably) some discrepancies between deployed contracts and the abi used.
Let's ignore this for now, since those contracts and tests mainly concern the stale  `rollup-keyper` design.

PR #433, opened 2024-02-21
Upgrade the WASM library

konradkonrad:
This PR prepares the `2.0.0` release of the `shutter-crypto` wasm library.

It
- builds the wasm library against `shlib 0.1.13`
- adds tests (with `jest`)
- runs the "json-tests" (from `./bin/rolling-shutter crypto testdata`)
- removes the restriction of the `epochId` parameter to be exactly `32 bytes`
- fixes a bug when calling `verifyDecryptionKey`
- changes the version string to `2.0.0`
- runs the wasm tests in CI


ulope:
LGTM.

Did linking the derived file not work in CI?
The idea was (and locally it works) that make would take care of updating the file if it’s stale (by doing its normal file timestamp magic).

ulope (approved):

konradkonrad:
> LGTM.
> 
> Did linking the derived file not work in CI? The idea was (and locally it works) that make would take care of updating the file if it’s stale (by doing its normal file timestamp magic).

I had plenty of iterations and at some point it _looked_ as if the linked file didn't resolve. I'm happy to retry at some later time, but for now I just wanted it to work.

PR #463, opened 2024-06-27
Fix duplicate keyper set error

jannikluhn:
On startup, the chainsyncer gives all keyper sets it can find in the keyper set manager to its subscribers, no matter if they are actually new or already in the db. This results in duplicate primary key errors when we try to insert them a second time.

To fix this, the SQL query now does not insert the keyper set if it is already present.


ulope (approved):
LGTM

PR #471, opened 2024-06-30
Downgrade "error dialing peer" log to debug

### LINKED ISSUE #467: Keypers spam "failed to dial" error logs
During normal operation, keypers show a lot of log messages similar to this:

```
keyper-1  | 2024/06/28 14:17:47.927391 ERR [           dht.go:96] error dialing peer error="failed to dial 12D3KooWCt78wkASE7KwtS4VenLoaPdxy9VmddzRC2UZKnRWPqEV: all dials failed\n  * [/ip4/127.0.0.1/tcp/23003] dial to self attempted\n  * [/ip4/172.19.0.5/tcp/23003] dial backoff" peer=12D3KooWCt78wkASE7KwtS4VenLoaPdxy9VmddzRC2UZKnRWPqEV
```

This should not be an error, but rather to be expected that an external address cannot be dialed. So this should be info or debug logs, ideally ones that are easily silencable.

We should also investigate why there's so many bad addresses in the DHT. It seems keypers advertize docker-internal addresses.

jannikluhn:
In production we see a lot of these messages. Failing to dial a peer is usually because the peer is unreachable which is not an error, so logging it as an error is misleading.

~Closes #467.~

### DIFF
--- a/rolling-shutter/p2p/dht.go
+++ b/rolling-shutter/p2p/dht.go
@@ -93,7 +93,10 @@ func findPeers(ctx context.Context, h host.Host, d discovery.Discoverer, ns stri
 				if h.Network().Connectedness(p.ID) != network.Connected {
 					_, err = h.Network().DialPeer(ctx, p.ID)
 					if err != nil {
-						log.Error().Err(err).Str("peer", p.ID.String()).Msg("error dialing peer")
+						log.Debug().
+							Err(err).
+							Str("peer", p.ID.String()).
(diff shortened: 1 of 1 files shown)


fredo:
As @konradkonrad pointed out, #467 also addresses the investigation about why docker internal addresses seem to be in the DHT. Please, either create a new issue for it as it is not part of this PR or adjust #467 and leave it open.

fredo (approved):

Issue #461, opened 2024-06-27
Fetch eon key publisher address from keyper set contract


jannikluhn:
To publish the eon key, the keypers use the eon key publisher contract. Each keyper set uses their own instance. The address is taken from a config file. This means that whenever we add a new keyper set, keypers who remain in the set have to update their config which is inconvenient.

To improve on this, the keyper should fetch the eon key publisher contract from the keyper set contract using the [getPublisher](https://github.com/shutter-network/shop-contracts/blob/9504b6278cefa1ef8a19208ec60ae5cf0e556460/src/KeyperSet.sol#L37-L39) function. The keyper set address in turn can be fetched from the keyper set manager.

fredo:
In addition to that we currently configure the EonKeyBroadcast contract which is not needed anymore since we publish the key via the EonKeyPublisher.

PR #478, opened 2024-07-02
Fix eon key publishing

### LINKED ISSUE #460: Eon key publishing error
Keypers fail to publish eon keys with the following error:

```
failed to check if eon key should be published error="no rows in result set" key=8a5d479286d9b0acec04c257ea670325f53214d9d8e940682b3dc94c5b4a1f5f0bc5bec62b52a345bf7088c6bb92df4008f46b86d72cd2baeb8548cf2ae56a610f7e1a42ceb2f868eeff75a9c2a0df1762d8085f57446532e398820b1e4a658e keyper-set-index=1
```

This is likely because of this line: https://github.com/shutter-network/rolling-shutter/blob/5fcb8af2fb870d7f9ecb3fa078048ec25676fdd8/rolling-shutter/eonkeypublisher/eonkeypublisher.go#L86

We use the key's keyper internal eon index instead of the keyper set index to get the corresponding keyper set from the db. If eon and keyper set index don't match, they get the wrong or in practice more likely no keyper set at all.

To fix this, we should query by keyper set index which is also part of the `EonPublicKey` event.

jannikluhn:
Closes #460 and #461

### DIFF
--- a/rolling-shutter/keyperimpl/gnosis/config.go
+++ b/rolling-shutter/keyperimpl/gnosis/config.go
@@ -166,7 +166,6 @@ func (c *GnosisConfig) TOMLWriteHeader(_ io.Writer) (int, error) {
 type GnosisContractsConfig struct {
 	KeyperSetManager     common.Address `shconfig:",required"`
 	KeyBroadcastContract common.Address `shconfig:",required"`
-	EonKeyPublish        common.Address `shconfig:",required"`
 	Sequencer            common.Address `shconfig:",required"`
 	ValidatorRegistry    common.Address `shconfig:",required"`
 }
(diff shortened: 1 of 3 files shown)


rolling-shutter/keyperimpl/gnosis/config.go
```diff
@@ -175,7 +174,6 @@ func NewGnosisContractsConfig() *GnosisContractsConfig {
 	return &GnosisContractsConfig{
 		KeyperSetManager:     common.Address{},
 		KeyBroadcastContract: common.Address{},
```
fredo, line 176: Do we need the `KeyBroadcastContract` still?

fredo (approved):
looks good to me

PR #481, opened 2024-07-02
Add pre-release key generation hack

jannikluhn:
This should be reverted some time next week.


fredo (approved):
I added the revert to the task list

PR #483, opened 2024-07-04
Add access node

jannikluhn:
This adds a new node type to the gnosis keyper network. It is a passive participant in the decryption keys gossip network. It is intended to be a peer for nodes that are unable of discovering peers dynamically to offload load from the keypers.

### DIFF
--- a/rolling-shutter/gnosisaccessnode/decryptionkeyshandler.go
+++ b/rolling-shutter/gnosisaccessnode/decryptionkeyshandler.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/shutter-network/shutter/shlib/shcrypto"
 
-	obskeyperdatabase "github.com/shutter-network/rolling-shutter/rolling-shutter/chainobserver/db/keyper"
 	"github.com/shutter-network/rolling-shutter/rolling-shutter/keyperimpl/gnosis"
 	"github.com/shutter-network/rolling-shutter/rolling-shutter/p2pmsg"
 )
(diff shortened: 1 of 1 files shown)


rolling-shutter/gnosisaccessnode/decryptionkeyshandler.go
```diff
@@ -0,0 +1,83 @@
...
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	"github.com/pkg/errors"
+
+	"github.com/shutter-network/rolling-shutter/rolling-shutter/p2pmsg"
+)
+
+const (
+	// TODO: pull these from keyper set manager
+	n         = 7
+	threshold = 5
```
fredo: threshold is 4

rolling-shutter/gnosisaccessnode/decryptionkeyshandler.go
```diff
@@ -39,25 +44,49 @@ func (handler *DecryptionKeysHandler) ValidateMessage(_ context.Context, msg p2p
...
 			handler.config.MaxNumKeysPerMessage,
 		)
 	}
+
+	eonKey, ok := handler.storage.GetEonKey(keys.Eon)
+	if !ok {
+		return pubsub.ValidationReject, errors.Errorf("no eon key found for eon %d", keys.Eon)
+	}
+
+	for i, k := range keys.Keys {
```
fredo, line 71: We should not forget to also apply optimizations here at one point, or ideally remove code redundancies.

fredo (approved):
looks good left some comments

PR #488, opened 2024-07-13
Use event index from contract instead of local counter

jannikluhn:


rolling-shutter/keyperimpl/gnosis/database/sql/queries/gnosiskeyper.sql
```diff
@@ -27,16 +34,9 @@ SET block_hash = $1, block_number = $2, slot = $3;
...
-INSERT INTO transaction_submitted_event_count (eon, event_count)
-VALUES ($1, $2)
-ON CONFLICT (eon) DO UPDATE
-SET event_count = $2;
-
 -- name: GetTransactionSubmittedEventCount :one
-SELECT event_count FROM transaction_submitted_event_count
-WHERE eon = $1
-LIMIT 1;
+SELECT max(index) + 1 FROM transaction_submitted_event
```
fredo, line 38: This should be true with the reorg resistance but could probably be false beforehand

rolling-shutter/keyperimpl/gnosis/newslot.go
```diff
@@ -219,7 +219,8 @@ func getTxPointer(ctx context.Context, db *pgxpool.Pool, eon int64, maxTxPointer
 			Int64("tx-pointer", txPointer).
 			Int64("tx-pointer-age", txPointerAge).
 			Msg("outdated tx pointer")
-		txPointer, err = gnosisKeyperDB.GetTransactionSubmittedEventCount(ctx, eon)
+		txPointerInt64, err := gnosisKeyperDB.GetTransactionSubmittedEventCount(ctx, eon)
```
fredo, line 222: the naming looks weird. Why is it called with `Int64` suffix, but then gets casted into int64

fredo (approved):
Left one nitpick but otherwise looks good

PR #490, opened 2024-07-13
Add basic reorg handling

jannikluhn:
Reorgs are detected by comparing each incoming header with the current sync head. If the incoming header is supposed to be the child of the sync head according to its block number, but the parent hash is different, a reorg is detected. Then, we delete the most recent events in the database.


fredo (approved):
lgtm

PR #501, opened 2024-07-22
Add libp2p seed node metrics

konradkonrad:
`p2pnode`, the `rolling-shutter` seed node for `libp2p`, did not publish metrics previously. This adds the `metricsserver` configuration during config parsing. 

I am unsure, whether this will need any extra steps, other than enabling `[Metrics] > enabled = true` in the config?

### DIFF
--- a/rolling-shutter/cmd/p2pnode/p2pnode.go
+++ b/rolling-shutter/cmd/p2pnode/p2pnode.go
@@ -3,7 +3,6 @@ package p2pnode
 import (
 	"context"
 
-	"github.com/pkg/errors"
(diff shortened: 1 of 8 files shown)


ulope (changes requested):
This still needs to actually start the metricsserver service in case metrics are enabled.

See here for example:
https://github.com/shutter-network/rolling-shutter/blob/6bab788a1c85ded24c8675575e4c0c0d12edf3fa/rolling-shutter/gnosisaccessnode/node.go#L55-L60

ulope (approved):
I added the instantiation of the MetricsServer

PR #510, opened 2024-07-24
libp2p add relay

ulope:
Enable the realy service on bootstrap nodes and enable automatic usage on other nodes


rolling-shutter/p2p/p2p.go
```diff
@@ -213,6 +216,17 @@ func createHost(
 			// Attempt to open ports using uPNP for NATed hosts.
 			libp2p.NATPortMap(),
 		)
+		if len(config.BootstrapPeers) > 0 {
+			options = append(options,
+				libp2p.EnableAutoRelayWithStaticRelays(config.BootstrapPeers),
```
ulope, line 221: We could possibly also use the access nodes here, but that would have been more work to gather the addresses.

I'd say lets test with this first and see if it changes anything.

konradkonrad (approved):
looks good to try!

PR #520, opened 2024-08-21
Cleanups

jannikluhn:
This PR
- removes a lot of code
- fixes formatting and naming conventions of the message protobufs
- makes identity preimage field naming consistent (renamed from epoch id and identity)


ulope (changes requested):
Very good to remove all that dead code.

I'm not clear on two things:
- ~Renaming the protobuf fields: Unless I'm misunderstanding this will break compatibility with existing code and also with the Nethermind clients.~ - Not an issue, Protobuf doesn't care about the field names on the wire protocol, just the tag number.
- ~Name changes from `GetInstanceID` to `GetInstanceId` (and a few similar ones): The golang styleguide says that `ID` (and similar abbreviations) should be in upper case.~ - This is caused by the field name changes in the protbuf messages. Apparently being golint compliant is not a goal of protoc (see golang/protobuf#53)

ulope (approved):

PR #522, opened 2024-08-22
floodsub discovery mechanism

blockchainluffy:
This PR contains new flood sub peer discovery message implementation. 
In this scope, it only publishes `Peer` message on the topic specified in config or on default topic and does not subscribe for listening to these updates. 
This implementation mimics the JS implementation - https://github.com/libp2p/js-libp2p-pubsub-peer-discovery/tree/main.

### DIFF
--- a/rolling-shutter/p2p/floodsubpeerdiscovery/gossippeerdiscovery.go
+++ b/rolling-shutter/p2p/floodsubpeerdiscovery/gossippeerdiscovery.go
@@ -8,8 +8,9 @@ import (
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/peerstore"
 	"github.com/rs/zerolog/log"
-	"github.com/shutter-network/rolling-shutter/rolling-shutter/medley/encodeable/address"
 	"google.golang.org/protobuf/proto"
+
+	"github.com/shutter-network/rolling-shutter/rolling-shutter/medley/encodeable/address"
 )
 
 const defaultTopic = "_peer-discovery._p2p._pubsub"


rolling-shutter/p2p/floodsubpeerdiscovery/gossippeerdiscovery.go
```diff
@@ -0,0 +1,113 @@
...
+	pd.PeerStore = config.PeerStore
+
+	if len(topics) > 0 {
+		for _, topic := range topics {
+			topic, err := pd.Pubsub.Join(topic)
+			if err != nil {
+				return fmt.Errorf("failed to join topic | err %v", err)
+			}
+			pd.Topics = append(pd.Topics, topic)
+		}
```
ulope, line 42: Is there ever a case where we need to join other topics besides the default?

rolling-shutter/p2p/floodsubpeerdiscovery/gossippeerdiscovery.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (pd *FloodsubPeerDiscovery) Start(ctx context.Context) error {
+	timer := time.NewTicker(time.Duration(pd.Interval) * time.Second)
+
+	for {
+		select {
+		case <-timer.C:
+			err := pd.broadcast()
+			if err != nil {
+				log.Info().Msgf("error in broadcasting floodsub msg | %v", err)
```
ulope, line 61: Should be a warning at least.

rolling-shutter/p2p/floodsubpeerdiscovery/gossippeerdiscovery.go
```diff
@@ -0,0 +1,113 @@
...
+		case <-ctx.Done():
+			return nil
+		}
+	}
+}
+
+func (pd *FloodsubPeerDiscovery) broadcast() error {
+	pubKey, err := pd.PeerId.ExtractPublicKey()
+	if err != nil {
+		return fmt.Errorf("peerId was missing public key | err %v", err)
```
ulope, line 73: With `fmt.Errorf` using `%w` instead of `%v` for errors makes more sense since it wraps the error and allows higher up code to inspect the underlying cause.

rolling-shutter/p2p/p2p.go
```diff
@@ -71,6 +73,13 @@ type p2pNodeConfig struct {
 	IsBootstrapNode    bool
 	IsAccessNode       bool
 	DiscoveryNamespace string
+	FloodsubDiscovery  FloodsubDiscoveryConfig
+}
+
+type FloodsubDiscoveryConfig struct {
+	Enabled  bool
+	Interval int
+	Topics   []string
```
ulope, line 82: Same question as above, do we ever need something else than the default topic?

rolling-shutter/p2p/config.go
```diff
@@ -46,8 +46,9 @@ type Config struct {
 	AdvertiseAddresses       []*address.P2PAddress `comment:"Optional, addresses to be advertised to other peers instead of auto-detected ones."`
 	CustomBootstrapAddresses []*address.P2PAddress `comment:"Overwrite p2p boostrap nodes"`
 	Environment              env.Environment
-	DiscoveryNamespace       string `shconfig:",required" comment:"Must be unique for each instance id."`
-	IsAccessNode             bool   `comment:"Optional, to be set to true if running an access node"`
+	DiscoveryNamespace       string                  `shconfig:",required" comment:"Must be unique for each instance id."`
+	IsAccessNode             bool                    `comment:"Optional, to be set to true if running an access node"`
+	FloodSubDiscovery        FloodsubDiscoveryConfig `shconfig:"required"`
```
ulope, line 51: There should be handling for `FloodSubDiscovery` in the `SetDefaultValues()` and `SetExampleValues()` methods below.

The question is if we should enable it by default.

ulope (changes requested):
LGTM in general, some small things below.

ulope (approved):
LGTM

PR #524, opened 2024-08-23
Simplify CI

jannikluhn:
This PR simplifies the CI workflows. Previously, the workflow was generated based on which parts of the code was changed in order to speed it up. Specifically, contracts, rolling-shutter, and end-to-end tests were enabled or disabled on demand.

However, since we don't use the contracts anymore and the end-to-end tests are outdated this is overkill. It also makes the CI configuration hard to maintain and edit. Therefore, this PR simplifies it. Now it just uses a static one that always runs the build, linting, and tests of rolling-shutter.

In order to make the CI workflow pass, this PR also
- disables the WASM build. It's currently failing, not used, and it's likely going to be removed
- adds missing skip statements for a few integration tests that should not be run as part of the unit test runs
- fixes two basic failing tests


ulope (approved):
LGTM except for the change in `shversion`.

PR #527, opened 2024-08-29
Use shutter-network/contracts/v2 bindings

### LINKED ISSUE #517: Merge contract repositories
Currently, we have two separate contract repositories: [shutter-network/shop-contracts](https://github.com/shutter-network/shop-contracts) and [shutter-network/gnosh-contracts](https://github.com/shutter-network/gnosh-contracts). We should merge them into one (e.g. `shutter-network/contracts`) and get rid of unused contracts. In particular:

- `shop-contracts`:
  - `KeyperSetManager`, `KeyperSet`, `KeyBroadcast` and `EonKeyPublish` is used in both the SHOP and the GNOSH implementation
  - `Inbox` is used in the SHOP implementation
- `gnosh-contracts`:
  - `KeyperSetManager`, `KeyperSet`, `KeyBroadcast` and `EonKeyPublish` are unused and can be removed
  - `Sequencer` and `ValidatorRegistry` are used in the GNOSH implementation
  - The interfaces `IKeyperSetManager`, `IKeyperSet`, `IKeybroadcast`, `IValidatorRegistry` and `ISequencer` are references to the GNOSH spec and should be preserved
- We have tests and scripts in both repositories. There is a probably a lot of overlap that we only need to preserve once. For the scripts, we should especially have two separate deploy scripts for shop and gnosh as well as a script to add a new keyper set.
- The two repos have automatically generated go bindings that are used as dependencies in rolling-shutter and other repos. They get generated in slightly different ways (see `gen/gen.go` in `shop-contracts` and `gen_bindings.sh` in `gnosh-contracts`). I personally prefer the latter, but others may disagree. We should end up with either a single or two separate packages. A single is probably preferable for simplicity.
- In `shop-contracts` there's some predeploy code that probably should be preserved, but I'm not entirely sure if we still need it.

konradkonrad:
This should fix #517

### DIFF
--- a/rolling-shutter/eonkeypublisher/eonkeypublisher.go
+++ b/rolling-shutter/eonkeypublisher/eonkeypublisher.go
@@ -13,7 +13,9 @@ import (
 	"github.com/jackc/pgx/v4/pgxpool"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
-	"github.com/shutter-network/shop-contracts/bindings"
+	"github.com/shutter-network/contracts/v2/bindings/eonkeypublish"
+	"github.com/shutter-network/contracts/v2/bindings/keyperset"
+	"github.com/shutter-network/contracts/v2/bindings/keypersetmanager"
 
 	obskeyperdb "github.com/shutter-network/rolling-shutter/rolling-shutter/chainobserver/db/keyper"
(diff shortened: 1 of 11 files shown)


konradkonrad:
Replacement for `github.com/shutter-network/shop-contracts/predeploy` still missing.

ulope (commented):
LGTM as far as I can tell, don't know if it actually works of course ;)

PR #535, opened 2024-10-14
feat: updated subscription to listen messages

blockchainluffy:


ulope (commented):
Since we now treat the floodsub messages like all other regular messages, wouldn't it make sense to include this into the `P2PMessaging` infrastructure? 
Seems a bit like reinventing a different wheel here.

konradkonrad:
> Since we now treat the floodsub messages like all other regular messages, wouldn't it make sense to include this into the `P2PMessaging` infrastructure? Seems a bit like reinventing a different wheel here.

I'd say it still makes sense to keep the topic specific stuff separate, but if all message handling is done the exact same way, then yes: once we have proven the concept, we should try not repeating us.

konradkonrad (approved):

PR #559, opened 2024-11-27
Add last event synched monitor

ylembachar:


rolling-shutter/keyperimpl/gnosis/syncmonitor.go
```diff
@@ -0,0 +1,61 @@
...
+	var lastBlockNumber int64
+	db := database.New(s.DBPool)
+
+	for {
+		select {
+		case <-time.After(checkInterval):
+			record, err := db.GetTransactionSubmittedEventsSyncedUntil(ctx)
+			if err != nil {
+				log.Warn().Err(err).Msg("error fetching block number")
+				continue
```
faheelsattar, line 41: I believe if the error is of type No rows found then you should continue, else for all other db type errors you should return it.

rolling-shutter/keyperimpl/gnosis/syncmonitor.go
```diff
@@ -0,0 +1,61 @@
...
+			} else {
+				log.Error().
+					Int64("last-block-number", lastBlockNumber).
+					Int64("current-block-number", currentBlockNumber).
+					Msg("block number has not increased between checks")
+				return errors.New("block number has not increased between checks")
+			}
+		case <-ctx.Done():
+			log.Info().Msg("stopping syncMonitor due to context cancellation")
+			return nil
```
faheelsattar, line 58: Also here just return `ctx.Err()`

PR #566, opened 2024-12-13
Test/newblock dec trigger

### LINKED ISSUE #563: [Shutter service] Tests for newblock / decryption trigger generation

faheelsattar:
Closes #563

### DIFF
--- a/rolling-shutter/keyperimpl/shutterservice/newblock.go
+++ b/rolling-shutter/keyperimpl/shutterservice/newblock.go
@@ -75,10 +75,14 @@ func (kpr *Keyper) shouldTriggerDecryption(
 ) bool {
 	nextBlock := triggeredBlock.Number.Int64()
 	keyperSet, err := obsDB.GetKeyperSet(ctx, nextBlock)
-	if err == pgx.ErrNoRows {
-		log.Info().
-			Int64("block-number", nextBlock).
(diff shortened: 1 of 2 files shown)


rolling-shutter/keyperimpl/shutterservice/newblock.go
```diff
@@ -56,6 +55,8 @@ func (kpr *Keyper) maybeTriggerDecryption(ctx context.Context, block *syncevent.
 		return errors.Wrap(err, "failed to query non decrypted identity registered events from db")
 	}
 
+	kpr.latestTriggeredTime = &block.Header.Time
+
```
faheelsattar, line 59: `kpr.latestTriggeredTime` should be updated after the db query is successfully executed. If we update it before the db query and for some reason there was an error while executing the query, we might miss some previous registered events since the `kpr.latestTriggeredTime` was already updated with new blocks timestamp.

faheelsattar:
`kpr.latestTriggeredTime` should be updated at the end after all the db reads/writes and error checks have been completed successfully. If we update it before there is a chance that we might end up missing some registered events since we already had updated the latest trigger time with the new block timestamp causing the decryption key to never get triggered for those registered events. I have moved the `kpr.latestTriggeredTime` at the end

PR #576, opened 2025-02-18
Upgrade golang to 1.23.6

ulope:
Also upgrade various codegen tools and remove obsolete ones


ulope:
Thanks for continuing on this @blockchainluffy.

LGTM, but Github won't let me approve (since it's my own PR 😂)

PR #607, opened 2025-06-17
feat: update isKeyper metric when handling messages and add successfull_dkg metric

blockchainluffy:


rolling-shutter/keyper/epochkghandler/key.go
```diff
@@ -52,25 +54,31 @@ func (handler *DecryptionKeyHandler) ValidateMessage(ctx context.Context, msg p2
 		return pubsub.ValidationReject, err
 	}
 	if !isKeyper {
+		keypermetrics.MetricsKeyperIsKeyper.WithLabelValues(strconv.FormatInt(eon, 10)).Set(0)
```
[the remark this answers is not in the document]
blockchainluffy: This is already set during DKG, but when the keyper restarts the metric loses its value. I have added it now to when keyper starts, to make it more consistent.

rolling-shutter/keyper/keypermetrics/metrics.go
```diff
@@ -79,7 +87,17 @@ var MetricsKeyperBatchConfigInfo = prometheus.NewGaugeVec(
 	},
 	[]string{"batch_config_index", "keyper_addresses"})
 
-func InitMetrics() {
+var MetricsKeyperSuccessfulDKG = prometheus.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Namespace: "shutter",
+		Subsystem: "keyper",
+		Name:      "successful_dkg",
```
ulope, line 94: Minor nitpick: I'd suggest to use `dkg_status` instead, that would allow to support multiple different statuses instead of just true/false

rolling-shutter/keyper/keypermetrics/metrics.go
```diff
@@ -88,4 +106,34 @@ func InitMetrics() {
 	prometheus.MustRegister(MetricsKeyperCurrentPhase)
 	prometheus.MustRegister(MetricsKeyperCurrentBatchConfigIndex)
 	prometheus.MustRegister(MetricsKeyperBatchConfigInfo)
+	prometheus.MustRegister(MetricsKeyperSuccessfulDKG)
+
+	queries := database.New(dbpool)
+	eons, err := queries.GetAllEons(context.Background())
+	if err != nil {
+		log.Error().Err(err).Msg("Failed to get all eons")
```
ulope, line 114: Mentioning that is from metrics in the error log would be nice.

ulope (approved):
LGTM

PR #640, opened 2025-10-22
Add Keyper count and threshold to batch config metric

ulope:


rolling-shutter/keyper/keypermetrics/metrics.go
```diff
@@ -199,7 +200,12 @@ func InitMetrics(dbpool *pgxpool.Pool, config kprconfig.Config) {
 
 			// Join keyper addresses for the label
 			keyperAddresses := strings.Join(batchConfig.Keypers, ",")
-			MetricsKeyperBatchConfigInfo.WithLabelValues(batchConfigIndexStr, keyperAddresses).Set(1)
+
+			keyperCount := strconv.Itoa(len(batchConfig.Keypers))
+
+			threshold := strconv.FormatInt(int64(batchConfig.Threshold), 10)
```
jannikluhn: any reason why you use `FormatInt` here, but in the line above `Itoa`?

PR #641, opened 2025-10-23
Add spec document for event trigger defintions

### LINKED ISSUE #637: Add rolling-shutter specs to include event-based decryption triggers

konradkonrad:
Fixes #637


jannikluhn (approved):
looks good, just two comments where the text could potentially be slightly clearer

PR #643, opened 2025-11-04
fix: decryption_key endpoint

blockchainluffy:
update findOperations to check for endpoint name for parameterised endpoints


jannikluhn:
https://github.com/shutter-network/rolling-shutter/blob/156dec4b2f293b94eb015e88982289c612d31bdd/rolling-shutter/keyper/kprapi/kprapi.go#L153-L155

We should switch the order of the middlewares around to protect the read only check from weird inputs.

PR #650, opened 2025-11-13
fix: enable multiple schema files to be parsed via sqlc definition

blockchainluffy:


rolling-shutter/medley/db/sqlc.go
```diff
@@ -48,32 +48,47 @@ func ParseSQLC(filesystem fs.FS, sqlcPath string) ([]Schema, error) {
...
-		// which is the path of the directory containing the schema files.
-		// This won't handle multiple directories, or individual sql file names.
-		schemaDirPath := path.Join(sqlcDir, entry.Schema)
-		dirEntries, err := fs.ReadDir(filesystem, schemaDirPath)
-		if err != nil {
-			return nil, errors.Wrap(err, "error reading schema dir")
+		var schemaPaths []string
+
+		// This allows only for schema directories, not individual sql files.
+		switch schema := entry.Schema.(type) {
```
jannikluhn: we should return an error if none of the types match

rolling-shutter/medley/db/sqlc.go
```diff
@@ -48,32 +48,47 @@ func ParseSQLC(filesystem fs.FS, sqlcPath string) ([]Schema, error) {
...
+		case string:
+			schemaPaths = []string{schema}
+		case []interface{}:
```
jannikluhn: why not cast directly to `[]string`?

rolling-shutter/medley/db/sqlc.go
```diff
@@ -48,32 +48,47 @@ func ParseSQLC(filesystem fs.FS, sqlcPath string) ([]Schema, error) {
...
+			for _, s := range schema {
+				if str, ok := s.(string); ok {
+					if str == "migrations" {
```
jannikluhn: can you add a comment explaining why the migrations dir is skipped? I think this might become hard to understand in the future.

PR #651, opened 2025-11-13
feat: enable keypers to update shuttermint keys

jannikluhn:
Keypers send check-in messages to the Shuttermint chain to announce their validator and encryption key. Instead of only accepting the first check in for each keyper, now Shuttermint accepts subsequent ones. The validator key will be used from the next block on. The encryption key will be emitted as a check in event.

To enable this and future Shuttermint updates, a fork heights struct has been added to the Shuttermint app state which specifies which protocol versions should go live at which block heights.

The keyper now stores all encryption keys and always uses the latest known one to encrypt directed messages. In addition, the encryption key cache in ShuttermintState has been removed as it was unused anyway.

### DIFF
--- a/rolling-shutter/app/types.go
+++ b/rolling-shutter/app/types.go
@@ -25,7 +25,7 @@ type GenesisAppState struct {
 
 func NewGenesisAppState(keypers []common.Address, threshold int, initialEon uint64, forkHeights *ForkHeights) GenesisAppState {
 	appState := GenesisAppState{
-		Threshold:   uint64(threshold),
+		Threshold:   uint64(threshold), //nolint:gosec // G115
 		InitialEon:  initialEon,
 		ForkHeights: forkHeights,
 	}


rolling-shutter/app/forks.go
```diff
@@ -0,0 +1,88 @@
...
+		CheckInUpdate: nil,
+	}
+}
+
+func (app *ShutterApp) IsCheckInUpdateForkActive() bool {
+	var override *ForkHeightOverride
+	if overrides, ok := forkHeightOverrides[app.ChainID]; ok {
+		override = overrides.CheckInUpdate
+	}
+	forkHeight := app.ForkHeights.CheckInUpdate
```
ylembachar: Maybe we should additionally check that ForkHeights is not nil here to avoid dereferencing, even though this wouldn't happen in the normal flow where the chain is initialized, but potentially in tests?

rolling-shutter/cmd/chain/init.go
```diff
@@ -73,6 +83,8 @@ func initCmd() *cobra.Command {
 	cmd.PersistentFlags().String("listen-address", "tcp://127.0.0.1:26657", "tendermint RPC listen address")
 	cmd.PersistentFlags().String("role", "validator", "tendermint node role (validator, isolated-validator, sentry, seed)")
 	cmd.PersistentFlags().Uint64("initial-eon", 0, "initial eon")
+	cmd.PersistentFlags().Int64("forks.check-in-update.height", 0, "block height at which to activate the check-in update fork")
+	cmd.PersistentFlags().Bool("forks.check-in-update.disabled", false, "whether the check-in update fork is disabled")
```
ulope: Double negatives can be confusing. Maybe `enabled` defaulting to `true` would be clearer.

PR #662, opened 2025-12-26
fix: update identity creation to include trigger definition

blockchainluffy:


konradkonrad (approved):
LGTM
-- crosslink https://github.com/shutter-network/shutter-api/pull/79/files#r2686081503

PR #672, opened 2026-02-24
fix: migrate event trigger keying to (eon, identity)

### LINKED ISSUE #673: Align event-trigger identity keying across shutter-api and rolling-shutter
(eon, identityPrefix, sender) is no longer a unique key after the identity computation change to include the trigger definition. See: https://github.com/shutter-network/rolling-shutter/pull/662/changes

Shutter API was updated to use (eon, identity) as the primary key, and this change mirrors that in the Keyper DB. See: https://github.com/shutter-network/shutter-api/pull/95

Example that exposed the issue: in the same eon, we registered triggers with the same identityPrefix and sender, but with different triggerDefinition. After identity computation started including triggerDefinition, these registrations produced different identities. With the old DB key (eon, identityPrefix, sender), they could be merged/overwritten at keying/upsert level, causing one trigger not to decrypt as expected.

ylembachar:


rolling-shutter/keyperimpl/shutterservice/config.go
```diff
@@ -82,7 +82,10 @@ func (c *Config) GetAddress() common.Address {
 }
 
 func (c *Config) EventBasedTriggersEnabled() bool {
-	return c.Chain.Contracts.ShutterEventTriggerRegistry != common.Address{}
+	return c != nil &&
+		c.Chain != nil &&
+		c.Chain.Contracts != nil &&
+		c.Chain.Contracts.ShutterEventTriggerRegistry != (common.Address{})
```
konradkonrad, line 88: I am not fully aware of the semantics of the parenthesis around `(common.Address{})` here?

konradkonrad (approved):
I have one question (see comments), but nothing to complain. LGTM!

rolling-shutter/keyperimpl/shutterservice/config.go
ylembachar: They're not needed here.

PR #677, opened 2026-03-12
Complex event data

konradkonrad:
There was an incorrect assumption when defining event trigger
matches on "complex" data values (i.e. those, spanning multiple data
words: bytes, string, ...).

Previously, we encoded the `Length` in the trigger definition and
expected to read that much from log data in order to match. However,
this would break, when the actual log data had a different alignment
(due to previous non-targeted arguments).

Instead we could remove `Length` from `LogValueRef`,
and introduce a `Dynamic` flag, to indicate fields for which we want to rely 
on the encoded length in the event data (see
https://docs.soliditylang.org/en/latest/abi-spec.html) at the specified
`Offset`.

We added more tests, that use a `SimulatedBackend` EVM to make sure, we
can properly decode and match actual EVM generated logs.

Since `Length` is no longer defined, we had to adjust some tests and
remove some edge case checks that no longer apply.

This increments the `Version` byte of event trigger definition RLP encoding to `0x02`.

Relates to https://github.com/shutter-network/shutter-api/issues/84
Critical for: https://github.com/shutter-network/shutter-api/issues/110

### DIFF
--- a/rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
+++ b/rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
@@ -15,7 +15,7 @@ import (
 
 const (
 	Word    = 32
-	Version = 0x1
+	Version = 0x2
 )
 
 // EventTriggerDefinition specifies an event-based trigger.
--- a/rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
+++ b/rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
@@ -309,7 +309,7 @@ func (r *LogValueRef) GetOffsetDataValue(log *types.Log) []byte {
 	return value
 }
 
-// aligns []byte to 32 byte
+// aligns []byte to 32 byte.
 func Align(val []byte) []byte {
 	words := (31 + len(val)) / Word
 	x := make([]byte, Word*words)
--- a/rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
+++ b/rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
@@ -448,11 +448,6 @@ func (p *ValuePredicate) validateArgValues() error {
 			return fmt.Errorf("integer argument %d cannot be negative for operation %d", i, p.Op)
 		}
 	}
-	// for i, arg := range p.ByteArgs {
-	// 	if len(arg) > Word {
-	// 		return fmt.Errorf("size of byte argument %d larger than 32 bytes for operation %d, got %d bytes", i, p.Op, len(arg))
-	// 	}
-	// }
 	return nil
 }
 
--- a/rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
+++ b/rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
@@ -39,7 +39,6 @@ type LogPredicate struct {
 //     byte 32, ends at byte 96 (exclusive), and is 64 bytes long.
 type LogValueRef struct {
 	Offset uint64
-	Length uint64
 }
 
 // ValuePredicate defines a condition on a value contained in an event log that must be satisfied
@@ -159,8 +158,10 @@ func (d *EventTriggerDefinition) ToFilterQuery() (ethereum.FilterQuery, error) {
 // Match checks if the log matches the event trigger definition by checking all log predicates.
 //
 // This may panic if Validate does not pass.
+// We need to match ABI encoding: https://docs.soliditylang.org/en/latest/abi-spec.html
 func (d *EventTriggerDefinition) Match(log *types.Log) (bool, error) {
 	if log.Address != d.Contract {
+		fmt.Println("Contract mismatch")
 		return false, nil
 	}
 	for _, logPredicate := range d.LogPredicates {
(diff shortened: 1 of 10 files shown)


rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
```diff
@@ -15,7 +15,7 @@ import (
 
 const (
 	Word    = 32
-	Version = 0x1
+	Version = 0x2
```
jannikluhn, line 18: If no one has used the old version yet I would just keep it as 0x1 to avoid potential confusion.

rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
```diff
@@ -159,8 +158,10 @@ func (d *EventTriggerDefinition) ToFilterQuery() (ethereum.FilterQuery, error) {
 // Match checks if the log matches the event trigger definition by checking all log predicates.
 //
 // This may panic if Validate does not pass.
+// We need to match ABI encoding: https://docs.soliditylang.org/en/latest/abi-spec.html
 func (d *EventTriggerDefinition) Match(log *types.Log) (bool, error) {
 	if log.Address != d.Contract {
+		fmt.Println("Contract mismatch")
```
jannikluhn: can be removed

rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
```diff
@@ -295,6 +268,55 @@ func (r *LogValueRef) GetValue(log *types.Log) []byte {
 	return value
 }
 
+// GetOffsetDataValue retrieves a "complex" data value from the log based on the LogValueRef.
+//
+// In case a slice of log data is referenced and the slice exceeds the log's data length, the
+// result will be zero-padded on the right to the expected length.
+func (r *LogValueRef) GetOffsetDataValue(log *types.Log) []byte {
```
jannikluhn: The idea of the `LogValueRef` type was to specify where the value can be found in the log. It seems like that's not the case anymore because there's now the `GetValue` an the `GetOffsetDataValue` functions and the caller needs to choose between the two based on outside info.

What do you think of adding a field `dynamic` or `complex` to `LogValueRef` to enable to make this distinction again? It will increase the serialized size by only one byte and is more explicit and clearer than extracting it from the operation (`p.ValuePredicate.Op != BytesEq`).

(Also, BytesEq was supposed to be used for bytes32 as well which is fixed sized, but I guess we could use integer equality for it instead, although that's a bit hacky).

rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
```diff
@@ -295,6 +268,55 @@ func (r *LogValueRef) GetValue(log *types.Log) []byte {
...
+	// abi encoded log data:
+	// W1: first argument value (simple) or offset_0 (complex)
+	// W2: second argument value (simple) or offset_1 (complex)
+	// W..@offset_0: size_0 for first complex argument
+	// W..@offset_0+W: start of data for first complex argument
+	//
+	// we need to discriminate between literal/simple and complex arguments
+	// - simple are found in `data[(offset-4)*WORD:(offset-4+1)*WORD]`
```
jannikluhn, line 257: can we add that this also applies to indexed complex arguments (which are hashed to a single word)? This confused me when if first read it.

rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
```diff
@@ -295,6 +268,55 @@ func (r *LogValueRef) GetValue(log *types.Log) []byte {
...
+		if endByte < availableEnd {
+			availableEnd = endByte
+		}
+		copy(value, log.Data[startByte:availableEnd])
+	}
+	return value
+}
+
+// aligns []byte to 32 byte.
+func Align(val []byte) []byte {
```
jannikluhn: it seems this is only used in the tests, so it should be defined in `eventtrigger_test.go`

rolling-shutter/keyperimpl/shutterservice/eventtrigger_test.go
```diff
@@ -2384,3 +2062,104 @@ func TestLogValueRefRLPRoundTrip(t *testing.T) {
...
+		}},
+	}
+	three := common.BytesToAddress(big.NewInt(84).Bytes())
+	mThree := LogPredicate{
+		LogValueRef:    LogValueRef{Offset: 3},
+		ValuePredicate: ValuePredicate{Op: BytesEq, ByteArgs: [][]byte{Align(three[:])}},
+	}
+	four := []byte("first and slightly longer arg that should use more space and if i am right, then this will span multiple words")
+	preFour := []byte("first and slightly longer arg that should use more space and if ")
+	mFoure := LogPredicate{
```
jannikluhn:
```suggestion
	mFour := LogPredicate{
```

typo?

rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
konradkonrad: There is at least one [demo application out there](https://onlydogs-chiado-825d4c9991b3.herokuapp.com/), that is incompatible after this change. Also, if anyone did consult the spec in the past and after this, they would get a hint, that something changed. My rationale was to rather use the version increments too often than too little - it also helps to get some practice on what needs to be done and what is painful. What do you think?

rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
konradkonrad: see this recent change: https://github.com/shutter-network/rolling-shutter/pull/677/changes#diff-6e726f9da3ffc61c7b94f209e2c9c961b9cf050f2392e77d442d45624a6420a1R284

konradkonrad: please resolve, if you're satisfied :)

rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
konradkonrad: See my proposal implemented in https://github.com/shutter-network/rolling-shutter/pull/677/changes/9261f79cb2f0afd6111bcdbe46cd0d75ba428363 and let me know, what you think!

rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
```diff
@@ -191,71 +194,48 @@ func (p *LogPredicate) Match(log *types.Log) (bool, error) {
...
-	if r.Length == 1 {
-		buf.WriteUint64(r.Offset)
-	} else {
-		listIndex := buf.List()
-		buf.WriteUint64(r.Offset)
-		buf.WriteUint64(r.Length)
-		buf.ListEnd(listIndex)
-	}
+	buf.WriteBool(r.Dynamic)
+	buf.WriteUint64(r.Offset)
```
jannikluhn: I think this shouldn't work, it needs a `buf.List()` call at the start to announce it's encoding an RLP list, not individual values. The same in the decoding function. However, since the encoding is static now (not either a list or a single value depending on the length), I don't think the custom `EncodeRLP` and `DecodeRLP` functions are necessary at all anymore. The RLP package should be able to encode/decode LogValueRef values out of the box. (missed this in my previous review)

jannikluhn (commented):
looks good, just one note about the RLP encoding/decoding

rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
jannikluhn: looks good

rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
jannikluhn: fine with me

rolling-shutter/keyperimpl/shutterservice/eventtrigger.go
konradkonrad: it did work, but I removed the custom code nevertheless ;)

PR #686, opened 2026-03-30
feat: add checks on batch config msg in keyper to not send invalid msg to shuttermint

### LINKED ISSUE #683: Malformed config 14 blocks all later transitions
In the context of the latest API Keyper set transition to eon 13, a malformed config 14 is prevents further transitions. Keypers only look up latest accepted batch config + 1, so because shuttermint never accepted config 14, the latest accepted config remains 13 and later transitions such as config 15 are never picked up.

blockchainluffy:
This PR is for v1.4.1 release

### DIFF
--- a/rolling-shutter/keyper/keyper.go
+++ b/rolling-shutter/keyper/keyper.go
@@ -337,13 +337,6 @@ func (kpr *KeyperCore) handleOnChainKeyperSetChanges(
 		return err
 	}
 
-	if nextKeyperConfigIndex == keyperSet.KeyperConfigIndex {
-		log.Debug().
-			Int64("keyper-config-index", keyperSet.KeyperConfigIndex).
-			Msg("batch config already sent (scheduled).")
-		return nil
-	}
-
 	keypers, err := shdb.DecodeAddresses(keyperSet.Keypers)
 	if err != nil {
 		return err
--- a/rolling-shutter/keyper/keyper.go
+++ b/rolling-shutter/keyper/keyper.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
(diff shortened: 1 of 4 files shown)


rolling-shutter/keyper/keyper.go
```diff
@@ -272,26 +316,40 @@ func (kpr *KeyperCore) handleOnChainKeyperSetChanges(
...
-	if lastSent == keyperSet.KeyperConfigIndex {
-		log.Debug().
+
+	if err := validateBatchConfig(latestBatchConfig, keyperSet, keypers); err != nil {
+		log.Warn().Err(err).
 			Int64("keyper-config-index", keyperSet.KeyperConfigIndex).
-			Msg("batch config already sent (scheduled).")
+			Msg("batch config validation failed, not sending to shuttermint")
+		// Mark as sent so we don't retry the same invalid config every cycle.
+		if err := q.SetLastBatchConfigSent(ctx, keyperSet.KeyperConfigIndex); err != nil {
```
ylembachar: `last_batch_config_sent` is being used here for configs that were not actually sent. That changes its meaning from “last config enqueued to Shuttermint” to “last config processed/skipped”, which makes the persisted state inaccurate and can cause confusion later. If the goal is to advance past invalid configs, this should use a separate cursor or be renamed end-to-end to match the new semantics if we do not need to preserve the distinction between “sent” and “skipped”.

rolling-shutter/keyper/keyper.go
```diff
@@ -272,26 +316,40 @@ func (kpr *KeyperCore) handleOnChainKeyperSetChanges(
 		return err
 	}
 	cq := obskeyper.New(tx)
-	keyperSet, err := cq.GetKeyperSetByKeyperConfigIndex(
-		ctx,
-		int64(latestBatchConfig.KeyperConfigIndex)+1,
-	)
+
+	lastSent, err := q.GetLastBatchConfigSent(ctx)
```
ylembachar: Similarly here, renaming `lastSent` to `lastHandledConfig` would make the flow clearer, since this value is no longer strictly “sent” but effectively the latest config already processed.

rolling-shutter/keyper/keyper.go
blockchainluffy: renamed lastBatchConfigSent to lastBatchConfigProcessed, not updating db schema

ylembachar (approved):
@blockchainluffy Looks good to me, provided that we align the DB name in an upcoming release if you could please create an issue for that on rolling shutter. Thank you.

PR #688, opened 2026-03-30
Ensure graceful handling of serialization version mismatches on event trigger definitions

### LINKED ISSUE #687: Handle version differences in shutter-api event trigger definitions
We need to make sure that changes in the serialization of event trigger definitions and/or version increments, do not lead to malfunction in keyper code.

Legacy entries that cannot be deserialized by current client code, need to be logged and discarded.

This issue is here to ensure we have handling code in place and unit tests to ensure it.

konradkonrad:
There are to my knowledge only two codepaths that unmarshal event trigger definitions. I've added an explicit version check that will log and skip any entries that do not match the currently implemented serialization version.

Fixes #687

### DIFF
--- a/rolling-shutter/keyperimpl/shutterservice/triggerprocessor.go
+++ b/rolling-shutter/keyperimpl/shutterservice/triggerprocessor.go
@@ -63,6 +63,13 @@ func (tp *TriggerProcessor) FetchEvents(ctx context.Context, start, end uint64) 
 			Logger()
 
 		trigger := EventTriggerDefinition{}
+		if len(triggerRegisteredEvent.Definition) > 0 &&
+			triggerRegisteredEvent.Definition[0] != Version {
+			triggerLog.Log().Int64("version", int64(
+				triggerRegisteredEvent.Definition[0])).Msg(
+				"ignoring trigger definition with outdated version in database")
+			continue
+		}
 		err := trigger.UnmarshalBytes(triggerRegisteredEvent.Definition)
 		if err != nil {
 			// This is not supposed to happen as only valid triggers are inserted into the database.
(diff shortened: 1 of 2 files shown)


rolling-shutter/keyperimpl/shutterservice/triggerprocessor.go
```diff
@@ -64,6 +64,13 @@ func (tp *TriggerProcessor) FetchEvents(ctx context.Context, start, end uint64)
 
 		trigger := EventTriggerDefinition{}
 		err := trigger.UnmarshalBytes(triggerRegisteredEvent.Definition)
+		if len(triggerRegisteredEvent.Definition) > 0 &&
```
ylembachar, line 66: To avoid unmarshaling a definition that will anyway be skipped, the version check should happen before `UnmarshalBytes()`, similarily to your previous update: https://github.com/shutter-network/rolling-shutter/pull/688/changes#diff-bf9d9e2e61cab2782870a4b2cf9bf3feae73dfe74d1265c1eec16652699a7d9dR95

rolling-shutter/keyperimpl/shutterservice/eventtrigger_test.go
```diff
@@ -2232,3 +2232,26 @@ func Align(val []byte) []byte {
 	copy(x[len(x)-len(val):], val)
 	return x
 }
+
+func TestInvalidVersionDoesNotCrash(t *testing.T) {
```
ylembachar: Also, I don’t think this test would be relevant once we check the version before unmarshalling, as that code path would be dead. A more relevant test for this PR would verify that outdated versions are skipped in the processor path instead.

PR #691, opened 2026-04-03
fix: identity registration events to use keyper-set-index

### LINKED ISSUE #692: Eon from identity registration events should be treated as keyperset index and use latest eon for the same keyperset index

blockchainluffy:
closes https://github.com/shutter-network/rolling-shutter/issues/692

### DIFF
--- a/rolling-shutter/keyperimpl/shutterservice/newblock.go
+++ b/rolling-shutter/keyperimpl/shutterservice/newblock.go
@@ -146,7 +146,7 @@ func (kpr *Keyper) resolveDecryptableEon(
 
 	_, isKeyper, err := coreKeyperDB.GetKeyperIndex(ctx, keyperConfigIndex, kpr.config.GetAddress())
 	if err != nil {
-		if err == pgx.ErrNoRows {
+		if errors.Is(err, pgx.ErrNoRows) {
 			log.Info().
 				Int64("keyper-set-index", keyperConfigIndex).
 				Msg("skipping event as keyper set has not been found")
--- a/rolling-shutter/keyperimpl/shutterservice/newblock.go
+++ b/rolling-shutter/keyperimpl/shutterservice/newblock.go
@@ -11,7 +11,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
 
-	obskeyper "github.com/shutter-network/rolling-shutter/rolling-shutter/chainobserver/db/keyper"
 	corekeyperdatabase "github.com/shutter-network/rolling-shutter/rolling-shutter/keyper/database"
 	"github.com/shutter-network/rolling-shutter/rolling-shutter/keyper/epochkghandler"
 	servicedatabase "github.com/shutter-network/rolling-shutter/rolling-shutter/keyperimpl/shutterservice/database"
@@ -84,7 +83,10 @@ func (kpr *Keyper) prepareTimeBasedTriggers(ctx context.Context, block *synceven
 	for _, event := range nonTriggeredEvents {
 		trigger, err := kpr.shouldTriggerDecryption(ctx, event, block)
 		if err != nil {
-			return nil, errors.Wrapf(err, "failed to check if should trigger decryption for event %d", event.Eon)
+			return nil, errors.Wrapf(err,
+				"failed to check if should trigger decryption for keyper set index %d",
+				event.Eon,
+			)
 		}
 		if trigger {
 			eventsToDecrypt = append(eventsToDecrypt, event)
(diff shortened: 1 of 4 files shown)


rolling-shutter/keyperimpl/shutterservice/newblock.go
```diff
@@ -202,30 +258,14 @@ func (kpr *Keyper) prepareEventBasedTriggers(ctx context.Context) ([]epochkghand
 		if len(firedTriggers) == 0 {
 			continue
 		}
-		eonStruct, err := coreKeyperDB.GetEon(ctx, eon)
+		eonStruct, decryptable, err := kpr.resolveDecryptableEon(ctx, coreKeyperDB, eon)
```
konradkonrad, line 261: signature is 
```golang
func (kpr *Keyper) resolveDecryptableEon(
    ctx context.Context,
	coreKeyperDB *corekeyperdatabase.Queries,
	keyperConfigIndex int64
) (corekeyperdatabase.Eon, bool, error) {…}
```
In other words: here we implicitly cast between `keyperConfigIndex` and `eon`.

konradkonrad: This is honestly puzzling to me, because `eon` is a value contained in `FiredTrigger` of service db and should not be a "keyperConfigIndex" kind? Or should not need resolution into a `corekeyperdatabase.Eon`?

Could it be that the triple return is actually superfluous, because the call does not "resolve" the `eon`, but only verifies "decryptability", given the current keyper and active keyper set?

konradkonrad: aha, we "resolve" eon properties such as `ActivationBlockNumber` that we need later on. So I guess we're resolving a type.

IOW: if we want to avoid the odd cast, we might want to consider changing the argument name to `eonNumber`, because the value we're handling is not strictly a ~`keyperSetIndex`~ `keyperConfigIndex` but rather the numeric for a specific `eon`, right?

rolling-shutter/keyperimpl/shutterservice/newblock.go
```diff
@@ -130,18 +121,76 @@ func (kpr *Keyper) shouldTriggerDecryption(
...
+			err,
+			"failed to get latest started eon for keyper set index %d",
+			keyperConfigIndex,
+		)
+	}
 
-	// don't trigger if we're not part of the keyper set
+	_, isKeyper, err := coreKeyperDB.GetKeyperIndex(ctx, keyperConfigIndex, kpr.config.GetAddress())
+	if err != nil {
+		if err == pgx.ErrNoRows {
```
ylembachar: This check won’t fire as intended as `GetKeyperIndex()` wraps the underlying `GetBatchConfig()` error, so when the batch config is missing the returned error is not `== pgx.ErrNoRows`. We should use `errors.Is(err,pgx.ErrNoRows)` here.

rolling-shutter/keyperimpl/shutterservice/newblock.go
ylembachar: I don’t think renaming the parameter to `eonNumber` would be correct here. This PR is changing the semantics of the even eon field. The incoming value is no longer being treated as a real core eon number; it is being treated as the submitted keyper-set/config index from the registry event, and `resolveDecryptableEon()` then resolves the latest eon record for that set. So the oddity is not the function signature, it’s that the service DB field is still named Eon even though this PR changes its semantics. If we want to reduce confusion, the cleaner follow-up is to rename the internal shutter-service representation at the ingestion boundary (e.g. keyperConfigIndex / submittedKeyperConfigIndex) and keep “eon” only for the resolved `corekeyperdatabase.Eon`. We should do the same on the API side as well: if the API is submitting a keyper-set/config index, we should name it that way there too, rather than continuing to call it eon and relying on implicit reinterpretation downstream.

blockchainluffy: I agree for renaming the eon field to keyper set index in such places, but this is not the scope of PR as the eon field in database for many tables like `decryption_key` and `decryption_key_shares` etc still correspond to keypersetindex. I believe, renaming should be part of a larger PR which handles all these mismatches

ulope (approved):
Only did a superficial review, but in that it looked good.

rolling-shutter/keyperimpl/shutterservice/newblock.go
ylembachar: Could you please open a follow-up issue for this refactor @blockchainluffy and resolve this comment? Thank you.

konradkonrad: thanks for the clarification, I think I got it now :)

PR #694, opened 2026-04-16
feat: check for priv_validator_state.json explcitly and create if does not exist

### LINKED ISSUE #695: priv_validator_state.json to be created if it does not exist
Currently, we only check for priv_validator_key.json file and assume that state file would exist.
The fix would check if state file also exists and create it if it does not with default values for `chain init` cmd.

blockchainluffy:
closes https://github.com/shutter-network/rolling-shutter/issues/695

### DIFF
--- a/rolling-shutter/cmd/chain/init.go
+++ b/rolling-shutter/cmd/chain/init.go
@@ -204,11 +204,20 @@ func initFilesWithConfig(tendermintConfig *cfg.Config, config *Config, appState 
 		privValStateFile := tendermintConfig.PrivValidatorStateFile()
 		var pv *privval.FilePV
 		if tmos.FileExists(privValKeyFile) {
-			pv = privval.LoadFilePV(privValKeyFile, privValStateFile)
-			log.Info().
-				Str("privValKeyFile", privValKeyFile).
-				Str("stateFile", privValStateFile).
-				Msg("Found private validator")
+			if tmos.FileExists(privValStateFile) {
+				pv = privval.LoadFilePV(privValKeyFile, privValStateFile)
+				log.Info().
+					Str("privValKeyFile", privValKeyFile).
+					Str("stateFile", privValStateFile).
+					Msg("Found private validator")
+			} else {
+				pv = privval.LoadFilePVEmptyState(privValKeyFile, privValStateFile)
+				pv.Save()
+				log.Info().
+					Str("privValKeyFile", privValKeyFile).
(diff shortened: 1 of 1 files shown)


rolling-shutter/cmd/chain/init.go
```diff
@@ -204,11 +204,20 @@ func initFilesWithConfig(tendermintConfig *cfg.Config, config *Config, appState
 		privValStateFile := tendermintConfig.PrivValidatorStateFile()
 		var pv *privval.FilePV
 		if tmos.FileExists(privValKeyFile) {
-			pv = privval.LoadFilePV(privValKeyFile, privValStateFile)
-			log.Info().
-				Str("privValKeyFile", privValKeyFile).
-				Str("stateFile", privValStateFile).
-				Msg("Found private validator")
+			if tmos.FileExists(privValStateFile) {
```
ylembachar, line 207: The goal of this PR it to allow operators rebuilding a node after a consensus failure to only preserve `priv_validator_key.json` and then resync, without manually reconstructing `priv_validator_state.json`.

The current implementation is not safe enough though. Right now it recreates `priv_validator_state.json` whenever the key exists and the state file is missing. That is too broad, because it also covers cases where old chain/app data is still present. In that situation, we would be silently resetting the validator signing state on a non-fresh node.

I think we should narrow the behavior:
- If `priv_validator_key.json` exists, `priv_validator_state.json` is missing, and the node’s local chain/app data has been wiped for a full resync, recreate `priv_validator_state.json` automatically.
- If `priv_validator_state.json` is missing but existing chain/app data is still present, fail fast with a clear operator error instead of silently recreating it.

That gives us the recovery flow we want for operators, without making validator-state resets implicit in every missing-file case. The runbook should also be explicit later: preserve `priv_validator_key.json`, wipe local chain/app state, then restart and resync.

blockchainluffy: - if the `priv_validator_state.json` is part of chain/app data, if the operator wipes only the state and preserves the data then the chain/app is not usable, in that case as well it need to resync from scratch, which is only possible with a new default `priv_validator_state.json`. So, imo, the change satisfies all the criteria.

ulope: Yeah I think if someone partially deletes things in their data directory they're on their own anyway.

ulope (approved):
LGTM

I think we can accept the slight risk that someone could bring their node into an undefined state if they only delete the priv validator state file.

Issue #698, opened 2026-04-30
Epic: Fix event-based decryption liveness and batching behavior


ylembachar:
### Summary

Event-based decryption currently batches multiple undecrypted fired identities into one decryption trigger / key-share / key-release flow.

This causes independent event identities to depend on each other. If one identity in the batch is missing shares, it can block key construction for other identities that already have enough shares. While service-level signatures are still present, batching also means Keyper must sign the exact same identity batch; if their local undecrypted views differ, signatures can split across different batch hashes and keys may not be released.

Event-based identities are independent, so batching should be treated only as a P2P optimization, not as a correctness requirement.

### Problem

There are two separate steps affected by batching:

1. Build the key

Building a decryption key should only require enough valid shares for that specific identity. It should not fail because another identity in the same event-based batch is missing shares.

2. Publish the key

Right now, publishing keys requires threshold Keyper to sign the same batch of identities. If keypers have different local views of what is still undecrypted, they can sign different identity batches and no batch reaches threshold signatures.

The signatures prove threshold Keyper agreed to publish that exact batch. For event-based triggers, batch-level approval does not seem necessary because identities are independent and keys can be verified cryptographically.

### Proposed Direction

For event-based triggers, stop batching identities.

one fired identity -> one decryption trigger -> one key-share message -> one key-release flow

This keeps the fix contained to the event-based implementation and avoids changing keyper core behavior, which treats batches atomically for other implementations such as Gnosis.

### Relation to #617

Related: https://github.com/shutter-network/rolling-shutter/issues/617

Once #617 is fully implemented, service-level batch signature issues should be reduced or removed. However, removing event-based batching is a simpler contained fix that does not need to wait for that rollout.

### Why not change the Keyper core first?

Keyper core is shared by multiple implementations. Some implementations may rely on atomic batch behavior.

For Gnosis, slot batches are intended to be atomic: either all selected transactions for the slot are decrypted or none are. Gnosis key-share messages also include signatures over the identities being decrypted, and keys messages include those signatures. If a keys message only contains a subset of the originally signed identities, those signatures would be invalid for the subset.

So this epic should not change Gnosis behavior or Keyper core semantics as the first step.

### Message Volume

Removing batching increases P2P messages:

current: 1 message with N identities
new: N messages with 1 identity each

We are not too worried initially because API-side rate limiting already bounds event trigger registration, but fired-trigger bursts/backlogs can still increase message volume. We should test this on Chiado and add metrics/logging where useful.

[...]

PR #700, opened 2026-05-05
Remove batching from event based decryption triggers

### LINKED ISSUE #699: Use one event identity per key-share/key-release message
Change event-based decryption so each fired event identity is processed independently, with one identity per
decryption trigger, key-share message, and key-release flow. This avoids one event identity missing shares blocking unrelated identities, and avoids batch-signature divergence while service-level signatures are still present.

Tradeoff: this increases P2P messages from 1 batch with N identities to N single-identity messages. Message volume will be tested separately on Chiado.

konradkonrad:
This removes the "batching" when signing identities for event based decryption triggers. It also adds a `mise` based integration test, `mise run test-event-decryption`, to ensure the correct behavior.

Drive-by fixes:
- I had to upgrade the `postgres` version from `.tool-versions`
- There was a linting error that needed to be ignored (overflow warning when casting a blocknumber to uint).

Context is issue #698 / #699

Fixes #699


mise-test-setup/mise-tasks/test-event-decryption
```diff
@@ -0,0 +1,322 @@
...
+        zero32,
+    )
+
+
+# ---------------------------------------------------------------------------
+# DB helpers
+# ---------------------------------------------------------------------------
+
+
+def query_db(sql: str, keyper_index: int = 0) -> str:
```
jannikluhn: this function seems rather pointless and confusing, because all it does it flip arguments

mise-test-setup/mise-tasks/test-event-decryption
```diff
@@ -0,0 +1,322 @@
...
+         / os.environ["ETHEREUM_CHAIN_ID"]
+         / "run-latest.json").read_text()
+    )
+    addr = utils.get_uups_proxy_address(broadcast, "ShutterEventTriggerRegistryV1")
+    if not addr:
+        raise SystemExit("ShutterEventTriggerRegistryV1 proxy not found in deployment artifacts")
+    return addr
+
+
+def get_helper_bytecode(data_dir: Path) -> str:
```
jannikluhn: do we have a deploy script in the contracts repo that we can call instead, maybe even in the `deploy` task along the other contracts? Side effect would be that we don't redeploy it for every single run.

mise-test-setup/mise-tasks/test-event-decryption
```diff
@@ -0,0 +1,322 @@
...
+            ")"
+        )
+        if exists == "t":
+            print(f"  Decryption key ready: identity_prefix={identity_prefix_hex} → epoch_id={identity}")
+            return identity
+
+        time.sleep(poll_interval)
+
+
+def assert_no_batching(
```
jannikluhn, line 18: I think this test is a bit too detailed and single purpose for an end to end test. Instead, I would only test that the expected decryption key(s) are produced (similar to what we do in `test-decryption`

mise-test-setup/mise-tasks/test-event-decryption
```diff
@@ -0,0 +1,322 @@
+#!/usr/bin/env -S uv run --script
```
jannikluhn, line 1: The test setup is meant for setting up a testing environment and repeatedly running manual tests, like requesting keys and checking that it behaves as expected. This looks more like a full automated e2e test. I think we should have those and that they should use the test setup, but they should not be part of it itself. The same applies to the already existing `test-decryption` (which is much smaller though).

To make it fit, I think this should be split up:
- the test contract should be deployed as part of the existing `deploy` task
- there should be a `submit-event-registration` task (not sure which parameters are needed)
- there should be something like a `trigger-event` task to trigger the earlier registered event (not sure how it should be identified)
- there should be a task to wait for the generated key, either with `wait-for-decryption-key` or a new task (again not sure which parameter is needed to identify the event)

This allows to run tests manually and repeatedly with more freedom, maybe even with different event parameters or trigger conditions.

mise-test-setup/mise-tasks/init-keyper-dbs
```diff
@@ -47,39 +48,101 @@ def public_table_count(database_name: str) -> int:
...
-                "--config",
-                "/config.toml",
-            ]
+        init_db(index)
+        continue
+
+    # DB exists with tables. Re-initialize if the stored address doesn't match the config.
+    config_addr = utils.keyper_address(config_path)
+    db_addr = db_ethereum_address(database_name)
+    if db_addr.lower() != config_addr.lower():
```
jannikluhn, line 144: this (and the other changes in this file) seem unnecessary, why would the address change? And if they change, silently redeploying doesn't seem like the right fix

mise-test-setup/mise-tasks/utils.py
```diff
@@ -115,3 +115,41 @@ def get_created_contract_address(
         if isinstance(address, str) and address:
             return address
     return None
+
+
+def get_uups_proxy_address(
```
jannikluhn, line 129: I think this function often returns `None` or `continue`s instead of raising an error when there's actually a broken/unexpected input file

jannikluhn (commented):
Fix itself looks good, but some comments on the tests

mise-test-setup/mise-tasks/test-event-decryption
konradkonrad: agreed

PR #707, opened 2026-05-22
Run DKG on main chain

jannikluhn:
Run the DKG on the Ethereum-like chain that hosts the rest of the contracts as well. Depends on https://github.com/shutter-network/contracts/pull/9

It has been tested locally with the mise test setup and the added e2e tests. Not ready to merge, but putting it here for early review. 

- [ ] Update dependencies once the contract PR is merged
- [ ] Clean up commit history
- [ ] Squash db migrations
- [ ] Maybe: Move contract syncing from keyperimpls to shared module (maybe dkg). Not done yet to not make changes to the existing keyper set contract syncing logic
- [ ] Maybe: Backfill DKG events on restarts. Currently, a keyper that goes offline during DKG and misses contract events will likely not agree on the DKG result.
- [ ] Maybe: Remove eon key syncing which now is pointless.
- [ ] Maybe: Consolidate existing db tables into the same schema so that JOIN queries can access both DKG, eon, and keyper sets (would slightly simplify DKG handler logic)
- [ ] More testing, in particular on a real network with latency to the chain and between keypers


rolling-shutter/keyperimpl/shutterservice/newkeyperset.go
```diff
@@ -46,11 +55,110 @@ func (kpr *Keyper) processNewKeyperSet(ctx context.Context, ev *syncevent.Keyper
...
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	// ECIES key registration runs at the same architectural level as
+	// HandleBlock — once per discovered keyper set, after the keyper set
+	// row is committed (MaybeRegisterECIESKey reads it). It is idempotent
+	// and a no-op for non-members.
+	return kpr.dkgMgr.MaybeRegisterECIESKey(ctx, keyperConfigIndex)
```
ylembachar: Is there a reason we register the ECIESKey here and not in the Gnosis implementation? Could this method be part of the Keyper Core if it's the same across implementations?

rolling-shutter/keyperimpl/gnosis/newdkgevent.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/shutter-network/rolling-shutter/rolling-shutter/medley"
+	syncevent "github.com/shutter-network/rolling-shutter/rolling-shutter/medley/chainsync/event"
+)
+
+// processNewDKGEvent stores a DKG Contract event in the appropriate per-type
+// table and writes a `dkg_result` row when the event is a success notification.
+// Events for a `keyper_config_index` whose DKG already succeeded locally are
+// silently ignored so re-emitted or retried-and-superseded messages do not
+// pollute the message tables.
+func (kpr *Keyper) processNewDKGEvent(ctx context.Context, ev syncevent.DKGEvent) error {
```
ylembachar, line 22: This function also seems to be duplicated across implementations, I think we should avoid duplication when possible.

rolling-shutter/keyperimpl/gnosis/newdkgevent.go
```diff
@@ -0,0 +1,209 @@
...
+	keyperSetIndex, retryCounter := dkgEventKeys(ev)
+	keyperConfigIndex, err := medley.Uint64ToInt64Safe(keyperSetIndex)
```
ylembachar: Should we make these two consistent, and only use KeyperSetIndex, or is there a reason for this conversion?

rolling-shutter/keyperimpl/gnosis/keyper.go
```diff
@@ -278,8 +281,8 @@ func (kpr *Keyper) processInputs(ctx context.Context) error {
 			err = kpr.processNewBlock(ctx, ev)
 		case ev := <-kpr.newKeyperSets:
 			err = kpr.processNewKeyperSet(ctx, ev)
-		case ev := <-kpr.newEonPublicKeys:
-			err = kpr.processNewEonPublicKey(ctx, ev)
+		case ev := <-kpr.newDKGEvents:
```
ylembachar, line 284: Should we guarantee DKG-related events are processed in a correct order, or retried if they are not?

For example:
- `KeyperSetAdded` must be stored before a `DealingEvent`, because `storeDealing` needs the keyper set row to map poly evals to receivers.
- `DKGSucceeded` needs earlier dealing/accusation/apology rows to rebuild the local result.

Right now, if processing errors, `processInputs` logs and continues, so the event is dropped instead of retried.

rolling-shutter/txsender/txsender.go
```diff
@@ -0,0 +1,349 @@
...
+		Value:     value,
+		Data:      row.Data,
+	})
+	signed, err := types.SignTx(tx, types.LatestSignerForChainID(s.chainID), s.cfg.PrivateKey)
+	if err != nil {
+		s.markFailed(ctx, row.ID, row.Label, errors.Wrap(err, "sign transaction"))
+		return
+	}
+
+	// Mark the row submitted BEFORE broadcasting. If the process crashes
```
ylembachar, line 238: Should there be a recovery path for submitted txs that never get a receipt?

With mark-before-send, a crash before `SendTransaction` can leave a row `submitted` with a nonce/hash the node never saw. Since `PendingNonceAt` only reflects chain/node state, a later row could reuse that nonce, while the original row keeps polling forever. This can make the outbox state inconsistent and complicate later nonce handling.

For DKG txs, this means a keyper may believe it already submitted a dealing/accusation/apology/finalization action even though the contract never saw it, which can cause the keyper to miss the DKG attempt. Maybe store/rebroadcast signed txs while the phase is still active?

rolling-shutter/keyperimpl/gnosis/newdkgevent.go
jannikluhn: Yes, totally agree. The issue is the chainsyncer is used by both keyper level events (syncing keyper sets and now dkg events) and application specific events (gnosis transaction submitted events etc), as well as new blocks. Extracting it to some common place (DKG or keyper) is not trivial, especially without breaking synchrony between the two event classes. I've created an issue [here](https://github.com/shutter-network/rolling-shutter/issues/710) and would prefer to do it in a separate PR, is that fine with you or do you think we should do it now?

rolling-shutter/keyperimpl/gnosis/keyper.go
jannikluhn: Regarding the order, I don't think the risk does not justify the added complexity. It's true that we don't technically have a guarantee on the exact order of events coming on the channel (they're from different event subscriptions and I haven't found any documentation that says they're ordered properly). But in practice events should be spaced far enough apart (at least one block, usually one phase) that I don't think this should be an issue.

If we do want to improve ordering, I think the only generic way to do it would be to delay events and check if we receive another event soon after that should overtake the first one. That's fine for DKG (and now that I think about it maybe not even that complex), but for Gnosis where we have strong latency constraints I don't think we should do this. Do you have something like this in mind, or some other solutiuon?

Regarding retries in case of failures, I think there's three options: Ignore (current solution), retry, or crash. The main failure case I can think of is a db issue (down, disk full), I think then crash and restart would probably make the most sense (at least if we had proper backfilling, see #711) and is easiest to implement.

All in all, I would say these are issues with the existing event syncing (that now become more prominent), so I would prefer to not solve them separately and not as part of this PR. What do you think?

rolling-shutter/keyperimpl/shutterservice/newkeyperset.go
jannikluhn: no reason, fixed (by duplication, see comment below)

rolling-shutter/keyperimpl/gnosis/newdkgevent.go
jannikluhn: No reason, fixed (along with getting rid of other usages of keyper config index that should be keyper set index for consistency)

rolling-shutter/keyperimpl/gnosis/keyper.go
ylembachar: I only meant this for DKG events, not as a generic ordering delay for all Gnosis events.

I thought about a small in-memory pending/retry queue for DKG events, but we can maybe revisit it if we actually encounter this ordering issue, and it may also be covered by a proper backfill/recovery path.

For now, we could add a code comment/TODO near `processInputs` noting that DKG event failures are currently logged and not retried, and link it to #711?

rolling-shutter/dkg/manager.go
```diff
@@ -0,0 +1,328 @@
...
+		Int64("retry-counter", retryCounter)
+	if hasResult {
+		event = event.
+			Uint64("keyper-count", resultNumKeypers).
+			Uint64("threshold", resultThreshold).
+			Int("accusation-count", accusationCount).
+			Int("apology-count", apologyCount)
+	}
+	event.Msg("DKG succeeded")
+	return queries.InsertDKGResult(ctx, corekeyperdb.InsertDKGResultParams{
```
blockchainluffy, line 493: success=true is hardcoded but pureBytes can be nil (when ComputeResult fails or when pure is nil), so we store a success row with no share. Downstream treats it as a real success and the keyper can't produce shares for this eon, with no retry. Can we only store success=true+nil for the non-member case, and otherwise skip or write success=false with the error?

rolling-shutter/txsender/txsender.go
```diff
@@ -0,0 +1,353 @@
...
+	gasEstimate, err := s.cfg.Client.EstimateGas(ctx, ethereum.CallMsg{
+		From:  s.address,
+		To:    &to,
+		Value: value,
+		Data:  row.Data,
+	})
+	if err != nil {
+		// Gas estimation failures usually mean the transaction would revert;
+		// record it as failed so an operator can inspect the calldata.
+		s.markFailed(ctx, row.ID, row.Label, errors.Wrap(err, "estimate gas"))
```
blockchainluffy: good to have: the errors here can be of transient nature like rate-limits if using RPC provider or if the tx is already in mempool, we can disqualify these errors and should not mark them as failed, as they can be retried

rolling-shutter/keyperimpl/gnosis/newkeyperset.go
```diff
@@ -45,11 +54,111 @@ func (kpr *Keyper) processNewKeyperSet(ctx context.Context, ev *syncevent.Keyper
...
+			// Eagerly insert an eons row so the DKG participation loop has
+			// somewhere to anchor when the activation block approaches.
+			// Existing rows are tolerated because the chainsync initial poll
+			// can re-deliver KeyperSetAdded events that were already processed.
+			if _, err := coredb.GetEon(ctx, keyperSetIndex); err == nil {
+				return nil
+			} else if !errors.Is(err, pgx.ErrNoRows) {
+				return errors.Wrap(err, "check existing eon row")
+			}
+			dkgContract, phaseLength, leadLength := kpr.fetchDKGParamsForKeyperSet(ctx, ev.Contract)
```
blockchainluffy: fetchDKGParamsForKeyperSet is called inside the open BeginFunc tx, so up to 5 RPC round-trips run while holding a pool connection and an uncommitted write. On any RPC error it returns NULLs which InsertEon persists and GetEon skips re-insert on redelivery so it never retries. A NULL row is then fatal every block downstream, so the keyper is permanently excluded from this set's DKG. Can we move the RPC out of the tx and retry on failure instead of writing NULL?

rolling-shutter/keyperimpl/gnosis/newkeyperset.go
```diff
@@ -45,11 +54,111 @@ func (kpr *Keyper) processNewKeyperSet(ctx context.Context, ev *syncevent.Keyper
...
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	// ECIES key registration runs at the same architectural level as
+	// HandleBlock — once per discovered keyper set, after the keyper set
+	// row is committed (MaybeRegisterECIESKey reads it). It is idempotent
+	// and a no-op for non-members.
+	return kpr.dkgMgr.MaybeRegisterECIESKey(ctx, keyperSetIndex)
```
blockchainluffy, line 119: nit pick: New keypers can only register their ECIES key after joining a set (newkeyperset.go:97, ecies.go:43-47), so if (activationBlock - addedBlock) - dkgLeadLength is smaller than the registerKey tx inclusion+indexing time, maybeDeal deals empty slots to the unregistered keyper (puredkg.go:318) and its first round degrades into false accusations; can we document this timing invariant and defer dealing to a receiver whose key is not yet cached instead of dealing an empty slot?

blockchainluffy (commented):
The shuttermint code is deleted in this PR but the Primev implementation uses it still, just a thing to keep in mind that this version will not support Primev

rolling-shutter/keyper/keypermetrics/metrics.go
```diff
@@ -58,16 +48,6 @@ var MetricsKeyperEonStartBlock = prometheus.NewGaugeVec(
 	[]string{"eon"},
 )
 
-var MetricsKeyperIsKeyper = prometheus.NewGaugeVec(
```
blockchainluffy, line 61: can we keep this metric and populate it from the database from keyperset table?

rolling-shutter/txsender/txsender.go
```diff
@@ -0,0 +1,391 @@
...
+		log.Error().Err(err).
+			Int64("id", row.ID).
+			Str("label", row.Label).
+			Str("tx-hash", signed.Hash().Hex()).
+			Msg("tx outbox: mark submitted failed; skipping send")
+		return
+	}
+
+	if err := s.cfg.Client.SendTransaction(ctx, signed); err != nil {
+		s.markFailedUnlessTransient(ctx, row.ID, row.Label, errors.Wrap(err, "send transaction"))
```
ylembachar: On a transient failure this leaves the row as submitted (it was marked before the send), and nothing retries submitted rows: processPending only picks up pending, and the receipt poll never finds a hash that was never broadcast. So the tx is silently lost and the "will retry" log doesn't hold on this path. Maybe reset the row to pending here, or store/rebroadcast the signed tx as per this earlier thread: https://github.com/shutter-network/rolling-shutter/pull/707#discussion_r3350181297

rolling-shutter/dkg/finalizing.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		log.Warn().Err(err).
+			Int64("keyper-set-index", keyperSetIndex).
+			Int64("retry-counter", retryCounter).
+			Msg("cannot compute DKG result; skipping success vote")
+		// Mark the phase as resolved with a NULL tx_outbox_id row so the
+		// warning above runs at most once per DKG Instance. The local
+		// puredkg state is stable once all on-chain messages have been
+		// indexed; re-running ComputeResult on every block is redundant.
+		if insertErr := queries.InsertDKGSentAction(ctx, corekeyperdb.InsertDKGSentActionParams{
```
ylembachar, line 70: Here, a failed ComputeResult at the first Finalizing block permanently forfeits this Keyper's success vote for that retry, even though the phase has blocks left in which missing data could arrive. You could maybe just return on a ComputeResult error so the next block retries.

rolling-shutter/txsender/txsender.go
```diff
@@ -0,0 +1,391 @@
...
+}
+
+// isTransient reports whether err is a known transient chain-client failure
+// that warrants leaving the outbox row for the next tick rather than terminally
+// failing it. It is intentionally a positive allowlist: unknown errors fall
+// through and are treated as terminal by callers. Other transient failures
+// (e.g. HTTP 5xx from the RPC endpoint) may exist and are not covered here;
+// extend the list if a class of transient error is observed terminally failing
+// rows in production.
+func isTransient(err error) bool {
```
ylembachar, line 342: Should we add 429 and 5xx errors here as well? More generally, since wrongly giving up loses a DKG action while wrongly retrying only costs free operations, it may be simpler to retry all errors by default and mark rows failed only on positive evidence: a revert receipt, or the phase window ending.

rolling-shutter/keyper/keyper.go
```diff
@@ -199,234 +169,3 @@ func (kpr *KeyperCore) getServices() []service.Service {
...
-
-// TODO: we need a better block syncing mechanism!
-// Also this is doing too much work sequentially in one routine.
-func (kpr *KeyperCore) operateShuttermint(ctx context.Context, _ service.Runner) error {
-	for {
-		syncBlockNumber, err := retry.FunctionCall(ctx, kpr.blockSyncClient.BlockNumber)
-		if err != nil {
-			return err
-		}
-		keypermetrics.MetricsKeyperCurrentBlockL1.Set(float64(syncBlockNumber))
```
ylembachar, line 410: Adding this here to follow-up on our conversation since we need this metric on the alerting side and the gauge will now export a constant 0, if you could re-wire it to the new block source, e.g. set it in the keyperimpl processNewBlock handlers or in dkg.Manager.HandleBlock, which receives every new block number anyway.

rolling-shutter/txsender/txsender.go
jannikluhn: done in https://github.com/shutter-network/rolling-shutter/pull/707/changes/5ace118e99ac95e0286cd4330ac40e77e7785de2 (reset to pending)

rolling-shutter/txsender/txsender.go
jannikluhn: done in https://github.com/shutter-network/rolling-shutter/pull/707/changes/bb8329984697c4ec84cc07e4660eb4b5d92fd071 (adding 429 and 5xx)

I don't think errors should be retried by default, at least not by the tx sender. There are a lot of potential cases to cover where we don't want to retry (the main case being "execution reverted", but also all other kinds of EVM errors like insufficient funds, gas usage that exceeds the block gas limit, etc), it is difficult to enumerate them all. Also, 

On phase window ending, the tx sender doesn't know about the specifics of the DKG contract, it's generic for all types of transactions, so it can't handle these cases.

rolling-shutter/dkg/finalizing.go
jannikluhn: ComputeResult only processes dealings, accusations, and apologies, which can't arrive during the finalization phase. If it fails in the first finalizing phase block it'll fail in any other as well (at least as long events and blocks are processed in the correct order). So I don't think retrying is necessary or beneficial.

ylembachar: The computation runs on the local DB mirror (buildPureDKG reads the poly/accusation/apology tables), not on chain state directly, and that mirror is filled asynchronously by the event syncer. So ComputeResult can fail simply because an apology that's already on chain hasn't been indexed locally yet, which resolves a block or two later. The fix is minimal: on the ComputeResult error path, just return without writing the sent-action marker. The marker's only reader is this function's own guard, so skipping it lets the next block re-run the (already per-block) replay once the mirror catches up, instead of permanently forfeiting the vote. It's still written on the success path, so no double-vote, and the phase window bounds the retries.

Also, should we factor in genuine DKG failure here, e.g. more disqualified dealers than the threshold tolerates to avoid leaking key material?

Follow-up, more broadly: this is one symptom of a wider gap. DKG events arrive via push subscriptions on a separate stream from block heads with no ordering between them, so the phase clock can outrun the events we've actually indexed. Shuttermint never had this because everything was one totally-ordered stream. The same desync is behind the lag-induced false accusations and the startup retry-counter gap. Perhaps the real fix is a DKG-specific block-ordered pull indexer, for each block N, getLogs(N) → write events → then advance the cursor and run HandleBlock(N), so everything up to N is present when we act. That's essentially #711, and it subsumes the individual finalize/accusing patches. Maybe not for this PR, just flagging that these point fixes are patching symptoms of the one architectural gap but still worth keeping for now.

rolling-shutter/keyper/keyper.go
ylembachar: Fixed here: https://github.com/shutter-network/rolling-shutter/pull/707/changes/4707d735a24eb640ee355486169bc59c20ddaea5

rolling-shutter/dkg/finalizing.go
jannikluhn: We agreed offline that the assumption underneath is that we receive the different events (blocks or logs) from the subscriptions "sufficiently" ordered. We're going to validate this assumption with a long running test, while keeping this code as is for now to avoid a big rewrite.

rolling-shutter/dkg/dealing_test.go
```diff
@@ -0,0 +1,407 @@
...
+//     submitDealing, and a dkg_sent_actions row marking the dealing action
+//     as enqueued. It must NOT write to the shared `dkg_poly_commitments`
+//     or `dkg_poly_evals` tables — those are populated exclusively by the
+//     chain syncer from indexed events.
+//  2. Second invocation is a no-op — no new rows are inserted in any of those
+//     tables, because the dkg_sent_actions row already exists.
+//
+//nolint:funlen // fixture-heavy integration test; the "first-call writes, second is no-op" narrative is clearer inline.
+func TestMaybeDealPersistsInitialStateAndIsIdempotent(t *testing.T) {
+	if testing.Short() {
```
ylembachar, line 85: CI: the DB-backed dkg/ and txsender/ integration tests never run in CI

These tests gate on testing.Short() (so they t.Skip in the unit pass — .circleci/config.yml:159 runs -short), but none of them are named *Integration, so the integration pass (.circleci/config.yml:169 runs -run Integration) never selects them — even though the executor provisions ROLLING_SHUTTER_TESTDB_URL and a Postgres container (line 43).

Net result: the reactor tests (dealing / accusing / apologizing / finalizing), plus build_puredkg, ecies, active_dkgs, manager, and the DB-backed txsender tests — ~35 test functions — execute in neither CI pass. The coverage exists on disk but is inert, including the regression tests for the retry-counter fix and the txsender transient-error fix.

Fix: rename them to TestXxxIntegration, which drops them into the integration pass that's already provisioned with a database.

This can be a follow-up PR rather than a blocker here, but please do open a tracking issue and mark it required before release so it's isn't lost.

mise-test-setup/e2e-tests/mise.toml
```diff
@@ -0,0 +1,20 @@
+# DKG e2e tests harness. Env vars and parent tasks (e.g. `clean`,
```
ylembachar, line 1: These e2e tests are solid and cover the real DKG behavior (transitions, offline-recovery). They're understandably too heavy for per-PR CI, but they're also not wired into any automated cadence either (no nightly/dedicated stage in .circleci/). Is the plan to run them manually before releases, or should there be a scheduled/pre-release job? Mostly want it documented so it's clear when they're expected to run, especially since the DB-backed Go integration tests also aren't running per-PR (see my other CI comment).

Same here, this can be a follow-up rather than a blocker, but please open a tracking issue and mark it required before release so it isn't lost. The resolution might just be documenting that these run manually before release (rather than adding a CI job) — either way, let's decide it deliberately.

rolling-shutter/keyper/keypermetrics/metrics.go
```diff
@@ -58,16 +58,6 @@ var MetricsKeyperEonStartBlock = prometheus.NewGaugeVec(
...
-	prometheus.GaugeOpts{
-		Namespace: "shutter",
-		Subsystem: "keyper",
-		Name:      "is_keyper",
-		Help:      "Is this node a Keyper in the respective batch config",
-	},
-	[]string{"batch_config_index"},
-)
-
 var MetricsKeyperCurrentPhase = prometheus.NewGaugeVec(
```
ylembachar, line 61: Alongside the `is_keyper`/`current_block_l1` metrics added here, a few DKG metrics were orphaned by the Shuttermint removal: `current_phase`, `dkg_messages_sent`, and `dkg_messages_received` are still registered but populated nowhere (the code that set them, in smobserver, was deleted), so they report 0 forever and dashboards silently show "no DKG activity." Suggest removing the dead registrations. 

Related: `current_eon`/`dkg_status` are set only at startup (keyper.go:127) and go stale after any runtime DKG. 

And one gap worth adding: nothing distinguishes `dkg_result.success=true` from "I have a usable share" (the nil-share failure mode), a `keyper_dkg_has_share{keyper_set_index} `gauge would let us alert on `is_keyper==1 `AND `has_share==0`.

If you can open a follow-up issue for this as well, as these are not blocking but thought I would flag them.

rolling-shutter/dkg/ecies.go
```diff
@@ -0,0 +1,95 @@
...
+		member, err := m.resolveMembership(keyperSet)
+		if err != nil {
+			if errors.Is(err, errNotMember) {
+				return nil
+			}
+			return err
+		}
+		ownIndex := member.ownIndex
+
+		exists, err := coreQueries.ExistsECIESKey(ctx, shdb.EncodeAddress(m.cfg.OwnAddress))
```
ylembachar, line 52: This only checks that an `ecies_keys` row exists for our address, not that the registered key still matches the one we hold. The registry allows overwrites, so a stale registration can persist. Maybe we should compare the two and log an error on mismatch?

Issue #721, opened 2026-08-17
Don't send DKG messages right at phase boundary


jannikluhn:
In the on-chain-dkg branch, DKG messages are sent as soon as the phase in which they are valid starts. In rare cases, this leads to errors: Even though the Ethereum client already sent us the `newHead` event, the immediately following `eth_estimateGas` call may still be processed under the previous block. This results in the call failing.

The easiest way to fix this is to delay sending the DKG message by a block.

PR #725, opened 2026-08-26
fix(dkg): defer dispatch by one block at phase boundaries

jannikluhn:
Gas estimation races the phase boundary: an RPC node can announce block N via newHead while eth_estimateGas still executes against the state of N-1. A message enqueued on the boundary block then reverts with WrongPhase during estimation, which txsender classifies as non-transient and marks failed. Since the dkg_sent_actions marker is written atomically with the outbox row, the keyper never retries the action and sits out the whole phase.

Add DispatchPhaseAt, which matches PhaseAt except on the first block of each phase window, and route processDKG through it. The contract window is unchanged, so with phaseLength L the remaining L-1 blocks are ample for inclusion.

Windows with phaseLength <= 2 are exempt: they have no block that is both past the boundary and still has an in-window successor, since a transaction triggered by block B is included at B+1 at the earliest.

### DIFF
--- a/rolling-shutter/dkg/manager.go
+++ b/rolling-shutter/dkg/manager.go
@@ -326,8 +326,8 @@ func (m *Manager) activeDKGs(ctx context.Context, blockNumber uint64) ([]corekey
 // acceptable — the maybe-function will see the stale snapshot for one block
 // and pick up the new state on the next dispatch.
 //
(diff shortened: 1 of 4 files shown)


blockchainluffy (approved):

jannikluhn:
Closes #721

